PR #4, opened 2025-04-07
Squin Dialect Address Analysis

johnzl-777:
This is an implementation of the address analysis pass for the squin dialect with unit tests added.

A few things to keep in mind:

- @weinbe58 has helped me identify a problem in kirin with the constant propagation pass handling multiple return values from the `Apply` statement in squin. He's helped me implement a simple workaround for now as a new method table but there's a more concrete solution [here](https://github.com/QuEraComputing/kirin/pull/358)
- I had move the DAG Analysis into squin but I'm not sure we want to keep that there going forward, I just needed to stop this (rather hard to figure out) circular import from happening when I had to add `from bloqade import squin` into the `impls.py`. 
- ~~I remember when @Roger-luo was helping me build this out there was a moment where I was trying to use `@wraps` on `Apply` but there was some problem we ran into that required me to hold off on wrapping the `squin.wire` dialect. I don't quite recall the specifics but I think it would be worth making an issue to keep track of it so I can go ahead and wrap things nicely when it's fixed (:~~
- There's also a stopgap `Complex` type that should be removed when upgrading to the newer version of Kirin which does have the proper type


weinbe58:
For the 'wraps' issue the problem is we do not support unpacking multiple return values when lowering from python to kirin. As such there is no way to support a python frontend for 'wire.Apply'.

johnzl-777:
@weinbe58 got it, I'll remove the attempt at wrapping I commented out. I assume this shouldn't be a problem anyways because it isn't our intent to let people construct entire circuits with the squin dialect

weinbe58:
Actually we don't have a rewrite from qubits to wire dialect @johnzl-777 so writing anything is going to be difficult.

@Roger-luo we either need to implement python lowering multiple return values or implement the rewrite pass from the qubits to the wire dialect. ideally we would have both since you can get slightly more optimal code using the wires especially with subroutine calls

PR #11, opened 2025-04-08
Squin Operator Sites Analysis

johnzl-777:


src/bloqade/squin/analysis/shape/impls.py
```diff
@@ -0,0 +1,68 @@
...
+        rhs = frame.get(stmt.rhs)
+
+        if isinstance(lhs, OpShape) and isinstance(rhs, OpShape):
+            lhs_size = lhs.size
+            rhs_size = rhs.size
+            # Sized trait implicitly enforces that
+            # all operators are square matrices,
+            # not sure if it's worth raising an exception here
+            # or just letting this propagate...
+            if lhs_size != rhs_size:
```
johnzl-777, line 39: I feel like this should be an exception instead of just propagating a `NoShape?` because if you allow `Mult` to handle things that are not the same size you can get something that's not really valid or representable with the existing `Sized` trait

johnzl-777:
Hi @Roger-luo I believe this is a reasonable implementation of the squin operator shape analysis. I understand you also mentioned a "propagation of unitarity" in #11 but from what I can tell this would require something similar to a const prop analysis + wrap const rewrite? As in I'd have something like a `UnitaryAnalysis` pass and then a rewrite to set the traits.

It just doesn't seem very idiomatic to me to try and kill two birds with one stone in analysis here. 

If the behavior you see in `squin_op_playground.py` is correct I can go ahead and make unit tests before this gets merged in 👍

Roger-luo (commented):
neat PR, just some minor revision needed, great work!

src/bloqade/squin/analysis/shape/analysis.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+class ShapeAnalysis(Forward[Shape]):
+
+    keys = ["op.shape"]
+    lattice = Shape
+
+    def initialize(self):
+        super().initialize()
+        return self
```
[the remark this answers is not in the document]
johnzl-777: Good catch! I just blindly carried this over from the `AddressAnalysis` 😅

src/bloqade/squin/op/traits.py
```diff
@@ -5,22 +5,22 @@
...
 @dataclass(frozen=True)
-class Sized(ir.StmtTrait):
+class Sites(ir.StmtTrait):
     data: int
 
 
 @dataclass(frozen=True)
-class HasSize(ir.StmtTrait):
-    """An operator with a `size` attribute."""
+class HasNSitesTrait(ir.StmtTrait):
```
Roger-luo, line 13: Just `HasNSites`? I think `Trait` is redundant - just want to save some hand service time as old people :-)

johnzl-777: I agree it's pretty redundant but the problem I ran into last night was the change you proposed for the type lattice (from `OpShape` to `HasNSites`) collides with this unless I'm careful about fully qualifying both of them in code. If that's okay I can go ahead and just do that but otherwise this was the (admittedly rough) solution I came up with 😅

Roger-luo: Ah! I would use a longer name for the lattice then! Cuz that's unlikely used much

johnzl-777:
@Roger-luo I did my best to implement the name changes you requested but also did my best to pick some names that were still clear enough that one can disambiguate between them.

I also added unit tests to try to cover all the proper codepaths (:

squin_op_playground.py
```diff
@@ -0,0 +1,59 @@
+from kirin import ir, types
```
Roger-luo, line 1: you wanna delete the playground?

Roger-luo (approved):
nice! I think the code is great to go! Just remmeber to cleanup your playground

squin_op_playground.py
johnzl-777: Yes! Thank you for catching that. I should probably do a local ignore for an `_playground.py` files moving forward

Roger-luo: I usually use a `main.py` as my playground it's in the `.gitignore` already IIRC

PR #145, opened 2025-04-14
fix incorrect squin wire address type

johnzl-777:
I caught this when I was reviewing some old code, admittedly embarrassing not to have caught earlier 😅


Roger-luo (commented):
all good can we have a test somehow catch this?

johnzl-777:
@Roger-luo your wish is granted 🪄 ✨

Roger-luo (approved):

Issue #25, opened 2025-03-19
rewrite from qubit dialect to wire dialect


Roger-luo:
implement a rewrite rule that uses analysis to track the dependencies of qubits and rewrite that into `wire`s. This is similar to CUDA Q's quake pass. How this can be implemented is by maintaining a virtual table inside the rewrite rule and update the table with new allocated wire and replace the previous wire, e.g

```python
# SSA corresponding to the QReg => SSA corresponding to the current Wire
qreg2wire: dict[ir.SSAValue, ir.SSAValue]
```

┆Issue is synchronized with this [Jira Task](https://quera.atlassian.net/browse/OT-16) by [Unito](https://www.unito.io)

weinbe58:
@Roger-luo I'm thinking that maybe this rewrite could be implemented using a code-gen pass instead? that way the current implementation of the rewrites would not need to have an entire frame. I think this is probably a more general issue of when to do a rewrite vs when to do a code generation of Kirin IR between dialects. 

cc: @kaihsin

kaihsin:
port some of the discussion here: I think in this scenario (where we convert one dialect group to another) we actually want codegen, but we are abusing similar() + rewrite just because it's very obnoxious to create/codegen another ir.Method

Roger-luo:
I think for squin -> stim it makes sense to generate a new Method, but for qubit -> wire it doesn't. Because qubit -> wire is a rewrite happens within each method, where `stim` is a target that you want to generate code towards.

I agree we don't have a good Kirin IR -> Kirin IR codegen framework.

weinbe58:
> I think for squin -> stim it makes sense to generate a new Method, but for qubit -> wire it doesn't. Because qubit -> wire is a rewrite that happens within each method, where stim is a target that you want to generate code towards.

I think I disagree wires require frames to work properly and at that point, if we refactor the rewrite infrastructure to have frames to handle this, we've reproduced something very close to code gen. I think rewrites should be reserved for optimizations within a dialect group while code gen is translating between dialect groups.

Roger-luo:
> I think rewrites should be reserved for optimizations within a dialect group while code gen is translating between dialect groups.

hmm, I think this is a good point. Let's use this convention.

weinbe58:
This is a point that @kaihsin brought up but in our 1-on-1 discussion just to give the credit where it is due 😄

kaihsin:
[...]

PR #148, opened 2025-04-14
squin to stim rewrite

johnzl-777:


src/bloqade/squin/analysis/rewrite/stim.py
```diff
@@ -0,0 +1,30 @@
...
+    # Somehow need to plug in Address and Sites
+    # into the SSAValue Hints field, which only accepts
+    # Attribute types
+
+    ## Could literally just plug in `ir.Attribute` into
+    ## the Address and Site lattices? 
+    ## Couldn't I just create my own attributes instead?
+
+    address_analysis: Dict[ir.SSAValue, Address]
+    op_site_analysis: Dict[ir.SSAValue, Sites]
```
Roger-luo, line 23: I think you would want to store these analysis results in the SSAValue hints. They are useful for other passes too.

johnzl-777: Yes! Just taking inspiration from the `WrapConst` rewrite, I'm just thinking over how to get `Address` and `Sites` to be `Attributes` because of the type `hints` can accept.

johnzl-777:
Hi @Roger-luo , I believe what you see here should be a decent first pass with me trying to work with a bit of a mismatch in terms of how squin can be mapped to the existing stim dialect. Feel free to let me know what you think would be acceptable additions to either dialect based on my observations below, I can gladly spin that out as a separate PR/RFC.

I haven't taken advantage of the site analysis results just yet but it should be very easy for me to implement, I just wanted to make sure the analyses and overall rewrite behavior were (mostly) correct.

-  Reset as well as MeasureAndReset in the squin.qubit dialect line up more nicely with the Stim dialect as opposed to the squin.wire dialect owing to the fact that squin.qubit can take a whole ilist of qubits while you can only take a single wire per statement in squin.wire.
  - I know there's the ongoing RFC for Measurement (#158 ) but I think it's also worth looking at how flexible Reset should be in both squin dialects because Stim supports Reset Z (which I default to) but you have RX and RY. Perhaps allowing for an operator from squin.op or string option to be passed in to specify this?
  - Measure and MeasureAndReset both can have a `result` that isn't present in the Stim dialect and means I can't "eliminate" the old squin statements because the `result` could be depended on down the line. My understanding is you follow up your measurements with detector annotations but I assume this is something the user should specify and might also need to be added to the squin dialect? 
  - MeasureAndReset seems to have a more native representation in Stim with options for MRZ, MRY, and MRX. Currently I have to do a bit of a clunky two-statement representation: MZ, then RZ
- Should there be a way in squin to say I want the adjoint of some of the default `squin.op`'s? I ask this because all `Gate` subclasses in `stim.gate` have this option

Roger-luo:
> I know there's the ongoing RFC for Measurement (https://github.com/QuEraComputing/bloqade-circuit/issues/158 ) but I think it's also worth looking at how flexible Reset should be in both squin dialects because Stim supports Reset Z (which I default to) but you have RX and RY. Perhaps allowing for an operator from squin.op or string option to be passed in to specify this?

Yeah, can you start an RFC issue for reset? I think reset needs some more thoughts.

I think for the first pass, we can just skip those doesn't match well - squin is more expressive so not everything can be compiled into stim.

Roger-luo:
xref #25 I realize we need a codegen for stim instead of a rewrite. And to make the UX better, we need a separate analysis pass to validate all possible incompatibility of stim when compiling from squin.

johnzl-777:
> xref #25 I realize we need a codegen for stim instead of a rewrite. And to make the UX better, we need a separate analysis pass to validate all possible incompatibility of stim when compiling from squin.

I think this partly resolves the ongoing thread between myself, @ChenZhao44 and @weinbe58 over handling the friction from the fact that `Measure` as well as `MeasureAndReset` in squin can have a result that can be depended on later down the line while stim has no such thing. 

I know @ChenZhao44 was advocating for just throwing an exception and killing the rewrite if it detected any form of squin measurement being depended on while @weinbe58 pointed out it would be more proper to just "let it slide" and say no rewrite happened but let emission choke on it instead. 

@weinbe58 also suggested potentially "manipulating" the squin measurement result if it's being returned to something like "None" @Roger-luo has pointed out this easily violates the original intent of the function (and I'm sure causes even more problems downstream). 

I support @Roger-luo 's direction here as well as the result of @kaihsin 's and @weinbe58 's conversations that codegen seems more sensible here.

Roger-luo:
so rewrites technically are no-raise because the rewrite rules happens only when we are sure about the equivalence. If it cannot perform the rewrite, it will not be applied to the code instead of throwing an exception. 

So does codegen - we want to be able to catch as many issues as we can before running any codegen or rewrite so we can have a nice error message. This is kinda true for all the possible codegen/rewrite because you rely on compile-time info to do these codegen/rewrite. I think it's always good practice to do around of validation to be certain on the entire program before mutating/generating the code that one intends to do.

johnzl-777:
Got it 👍 I'll open up an issue to sketch out/track progress on the analysis passes to catch Stim-incompatible squin programs

johnzl-777:
I believe this rewrite now accepts Broadcast as well as changes from #158. This will probably have to change once #265 goes through but I don't think it will be too drastic a change.

There are a couple things (ordered in what I believe is the correct priority) I think need to be done as well from my notes in talking with @ChenZhao44 
- The noise statements in squin also need to map to the stim statements (I could probably knock this out in a day)
- There's also the matter of another QEC dialect that should be rewritten to stim as well that sits inbetween Chen's logical work
- There are still cases (maybe this is resolvable through another pass of sorts) where running DCE, because of potential dependencies from a lot of wire statements results, means the resulting stim rewrite from squin has dangling chunks/statements. I think some of my comments highlight this (or some of the unit tests too)
- Unit tests should of course, be made more robust (much of what is currently present is just "by eye" testing of output) but considering how much stuff is changing and being added, along with time pressure this will have to be deprioritized for now.

johnzl-777:
The stim codegen keeps complaining about duplicate registered methods for lowering constants?
Even when I do get it to work (I add the stim subdialects instead of `stim` in its entirety to the dialect group) then the output comes out very silly, something like: 

```
H constant constant constant
Z constant constant
```
I also realize I could definitely split out more of the rewrite logic into their own rules considering how much duplicate logic there is (with just a small variation required to handle address type extraction and wire routing). Probably worth a separate PR but in the interest of time I won't do it here (unless indicated otherwise!)

src/bloqade/squin/rewrite/wrap_analysis.py
```diff
@@ -0,0 +1,73 @@
...
+from kirin import ir
+from kirin.rewrite.abc import RewriteRule, RewriteResult
+from kirin.print.printer import Printer
+
+from bloqade.squin import op, wire
+from bloqade.analysis.address import Address
+from bloqade.squin.analysis.nsites import Sites
+
+
+@wire.dialect.register
```
kaihsin, line 13: is this only register to wire? or qubit dialect also need this? @Roger-luo

src/bloqade/stim/dialects/auxiliary/stmts/annotate.py
```diff
@@ -45,3 +45,7 @@ class NewPauliString(ir.Statement):
     flipped: tuple[ir.SSAValue, ...] = info.argument(types.Bool)
     targets: tuple[ir.SSAValue, ...] = info.argument(types.Int)
     result: ir.ResultValue = info.result(type=PauliStringType)
+
+
+# dialect_group = squin_extended = squin in bloqade-circuit + physical in bloqade-qec
+# Chen will need squin -> Stim rewrite IF it is a subroutine of rewrite from bloqade-qec extension
```
kaihsin, line 51: remove this

kaihsin (approved):
leave some comments. The question I have is which dialect that attribute should go to, but I don't think its a blocker.

src/bloqade/squin/rewrite/wrap_analysis.py
```diff
@@ -0,0 +1,73 @@
...
+
+    def print_impl(self, printer: Printer) -> None:
+        # Can return to implementing this later
+        printer.print(self.sites)
+
+
+@dataclass
+class WrapSquinAnalysis(RewriteRule):
+
+    address_analysis: Dict[ir.SSAValue, Address]
```
Roger-luo, line 46:
```suggestion
    address_analysis: dict[ir.SSAValue, Address]
```

src/bloqade/squin/rewrite/wrap_analysis.py
johnzl-777: This makes sense to me considering wire's Apply, Broadcast, and MeasureAndReset will return things whereas qubit doesn't. I haven't run into a case where the analysis fails on the qubit dialect.

PR #151, opened 2025-04-15
Forward `defs` from `qasm2.inline` to python Lowering. 

### LINKED ISSUE #150: inline QASM2 is not sharing the scope with python code
Before upgrading to Kirin 0.16  the following was OK but now it doesn't work anymore:

```
from bloqade import qasm2

def import_qasm(qasm2_prog):
    @qasm2.main.add(qasm2.inline_)
    def entry():
        qasm2.inline(qasm2_prog)
        
        return q
    return entry


test_qasm = """// Generated from Cirq v1.4.1 
OPENQASM 2.0;
include "qelib1.inc";
// Qubits: [q_0, q_1, q_2, q_3, q_4, q_5]
qreg q[6];
u3(pi*0.9999896015,pi*1.8867094803,pi*0.1132905197) q[2];
"""


kernel = import_qasm(test_qasm)

```

but how raises a BuildError:
```
  BuildError: q is not defined
    │  qasm2.inline(qasm2_prog)
    │  
  8│  return q
    │              ^
```

weinbe58:
in order to close #150 I added a flag that allows you to obtrain the `defs` from the last frame used in qasm2 lowering. The idea is those defs need to be given to the python lowering frame to make sure those names are availible to continue lowering the rest of the python code. This is OK because all declarations in qasm2 are global.

### DIFF
--- a/src/bloqade/squin/analysis/nsites/analysis.py
+++ b/src/bloqade/squin/analysis/nsites/analysis.py
@@ -15,9 +15,7 @@ class NSitesAnalysis(Forward[Sites]):
     keys = ["op.nsites"]
     lattice = Sites
 
-    # Take a page from const prop in Kirin,
-    # I can get the data I want from the SizedTrait
-    # and go from there
+    # Take a page from how constprop works in Kirin
 
     ## This gets called before the registry look up
     def eval_stmt(self, frame: ForwardFrame, stmt: ir.Statement):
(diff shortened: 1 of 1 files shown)


Roger-luo:
while this fixes #150 I don't think this is bug that should be fixed.

src/bloqade/qasm2/parse/lowering.py
```diff
@@ -27,6 +27,7 @@ def run(
         lineno_offset: int = 0,
         col_offset: int = 0,
         compactify: bool = True,
+        return_defs: bool = False,
```
Roger-luo, line 30: this is breaking the interface too. pyright would have errored here for you locally

PR #152, opened 2025-04-15
Implement pyqrack interpreter method for qasm2 glob.u gate

### LINKED ISSUE #15: Missing runtime implementation for `glob` dialect
The simulator is missing the concrete interpreter methods for the global gate dialect. A reference semantically global gates act on a list of registers, not qubits, hence one would need to loop over both the registers and then the qubits within the register, making sure to not act on qubits that have been lost.

david-pl:
Closes #15

### DIFF
--- a/src/bloqade/pyqrack/qasm2/core.py
+++ b/src/bloqade/pyqrack/qasm2/core.py
@@ -1,4 +1,7 @@
+from typing import Any
+
 from kirin import interp
+from kirin.dialects import ilist
 
 from bloqade.pyqrack.reg import (
     CBitRef,
@@ -9,7 +12,7 @@ from bloqade.pyqrack.reg import (
     PyQrackQubit,
 )
 from bloqade.pyqrack.base import PyQrackInterpreter
-from bloqade.qasm2.dialects import core
+from bloqade.qasm2.dialects import core, glob
 
 
 @core.dialect.register(key="pyqrack")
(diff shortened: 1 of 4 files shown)


src/bloqade/pyqrack/reg.py
```diff
@@ -33,7 +33,7 @@ class CBitRef:
     pos: int
     """The position of this bit in the classical register."""
 
-    def set_value(self, value: bool):
+    def set_value(self, value: Measurement):
```
david-pl, line 36: @weinbe58 this change here is just to make pyright happy, it's completely unrelated to the rest of the PR though (separate commit). Please double-check this.

src/bloqade/pyqrack/qasm2/core.py
```diff
@@ -77,3 +80,18 @@ def creg_eq(
             return (False,)
 
         return (all(left is right for left, right in zip(lhs, rhs)),)
+
+    @interp.impl(glob.UGate)
+    def ugate(self, interp: PyQrackInterpreter, frame: interp.Frame, stmt: glob.UGate):
```
weinbe58, line 85: This method needs to be implemented in a method table registered by `glob.dialect`. The convention is to put a method table inside a module with the same name as the dialect.

src/bloqade/pyqrack/qasm2/core.py
```diff
@@ -77,3 +80,18 @@ def creg_eq(
...
+        registers: ilist.IList[PyQrackReg, Any] = frame.get(stmt.registers)
+        theta, phi, lam = (
+            frame.get(stmt.theta),
+            frame.get(stmt.phi),
+            frame.get(stmt.lam),
+        )
+
+        for qreg in registers:
+            for qarg in qreg:
```
weinbe58, line 94: You can also implement the `__iter__` magic method for `PyQrackReg` to get the correct behavior here.

weinbe58 (approved):
left some minor comments.

Roger-luo (approved):

PR #159, opened 2025-04-15
`load` + `loadfile` methods for `QASM2` lowering. 

weinbe58:
addresses #150

### DIFF
--- a/src/bloqade/qasm2/parse/lowering.py
+++ b/src/bloqade/qasm2/parse/lowering.py
@@ -1,8 +1,11 @@
+from io import TextIOBase
+import os
+import pathlib
 from typing import Any
 from dataclasses import field, dataclass
 
 from kirin import ir, types, lowering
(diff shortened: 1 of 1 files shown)


src/bloqade/qasm2/parse/lowering.py
```diff
@@ -17,6 +19,118 @@ class QASM2(lowering.LoweringABC[ast.Node]):
     hint_show_lineno: bool = field(default=True, kw_only=True)
     stacktrace: bool = field(default=True, kw_only=True)
 
+    def loads(
+        self,
+        source: str,
+        kernel_name: str,
+        *,
+        returns: str | None = None,
```
weinbe58, line 27: We can add multiple returns later but for standard qasm2 file you can't do this because we do not support tuples.

Roger-luo: yeah I think this is ok.

PR #163, opened 2025-04-16
Fix some pyright issues

david-pl:
Got it down to 14 errors (starting out with 69). Not sure if all changes make sense.

The errors that are left are mostly ones such as this:

```python
kirin-workspace/bloqade-circuit/src/bloqade/stim/emit/stim.py:35:9 - error: Method "emit_block" overrides class "EmitABC" in an incompatible manner
    Return type mismatch: base method returns type "str", override returns type "str | None"
      Type "str | None" is not assignable to type "str"
        "None" is not assignable to "str" (reportIncompatibleMethodOverride)
```

I'm not sure what the best way to deal with those method overrides is.

Also, there's still a bunch of errors in `visual/animation`, which I could fix, but it would require some type assertions here and there (i.e. actually raising an error if the type expectation is not met) and I didn't want to go and break @kaihsin animation scripts.

We can discuss & continue here, or just merge (if the changes are fine) and continue in a separate PR, since this already improves things a bit.

### DIFF
--- a/src/bloqade/analysis/address/impls.py
+++ b/src/bloqade/analysis/address/impls.py
@@ -191,12 +191,11 @@ class SquinWireMethodTable(interp.MethodTable):
     ):
 
         origin_qubit = frame.get(stmt.qubit)
-        
-        if isintance(origin_qubit, AddressQubit):
 
+        if isinstance(origin_qubit, AddressQubit):
             return (AddressWire(origin_qubit=origin_qubit),)
         else:
-            return (Address.top(), )
+            return (Address.top(),)
 
     @interp.impl(squin.wire.Apply)
     def apply(
--- a/src/bloqade/analysis/address/impls.py
+++ b/src/bloqade/analysis/address/impls.py
@@ -190,9 +190,13 @@ class SquinWireMethodTable(interp.MethodTable):
         stmt: squin.wire.Unwrap,
     ):
 
-        origin_qubit = frame.get_casted(stmt.qubit, AddressQubit)
+        origin_qubit = frame.get(stmt.qubit)
+        
+        if isintance(origin_qubit, AddressQubit):
 
-        return (AddressWire(origin_qubit=origin_qubit),)
+            return (AddressWire(origin_qubit=origin_qubit),)
+        else:
+            return (Address.top(), )
 
     @interp.impl(squin.wire.Apply)
     def apply(
--- a/src/bloqade/analysis/address/impls.py
+++ b/src/bloqade/analysis/address/impls.py
@@ -204,17 +204,7 @@ class SquinWireMethodTable(interp.MethodTable):
         frame: ForwardFrame[Address],
         stmt: squin.wire.Apply,
     ):
-
-        origin_qubits = tuple(
-            [
-                frame.get_casted(input_elem, AddressWire).origin_qubit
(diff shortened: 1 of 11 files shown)


kaihsin:
For the CU3 bug, would you mind also adding unit tests for one of the angle to be zero?

david-pl:
Oh, CI fails because I added a method `get_present_trait` to kirin locally (just hitting F12 in vscode can do that). I'll open another PR there, because I think it makes sense to have it. But I we can also use `cast` here to assert types for the linter.

src/bloqade/analysis/address/impls.py
```diff
@@ -190,7 +190,7 @@ def unwrap(
         stmt: squin.wire.Unwrap,
     ):
 
-        origin_qubit = frame.get(stmt.qubit)
+        origin_qubit = frame.get_casted(stmt.qubit, AddressQubit)
 
         return (AddressWire(origin_qubit=origin_qubit),)
```
weinbe58, line 195:
```suggestion
        origin_qubit = frame.get(stmt.qubit)
        
        if isintance(origin_qubit, AddressQubit):

            return (AddressWire(origin_qubit=origin_qubit),)
        else:
            return (Address.top(), )
```

I think the implementation of this function is wrong maybe @johnzl-777 can confirm if this makes sense?

src/bloqade/analysis/address/impls.py
```diff
@@ -203,7 +203,10 @@ def apply(
...
 
         origin_qubits = tuple(
-            [frame.get(input_elem).origin_qubit for input_elem in stmt.inputs]
+            [
+                frame.get_casted(input_elem, AddressWire).origin_qubit
+                for input_elem in stmt.inputs
+            ]
         )
         new_address_wires = tuple(
             [AddressWire(origin_qubit=origin_qubit) for origin_qubit in origin_qubits]
```
weinbe58, line 212: This is also wrong you should just return all the values from the frame because there is no dependency on what the value is. 

```
return frame.get_values(stmt.inputs)
```

weinbe58 (commented):
LGTM but I think we might have some issues with the implementation in some areas not just linting.

src/bloqade/analysis/address/impls.py
johnzl-777: Yes, it does! Thank you for catching this (and explaining how to handle these pyright complaints moving forward 👍 )

src/bloqade/analysis/address/impls.py
david-pl: @johnzl-777 how about this one?

weinbe58: I talked to @johnzl-777 in person, he is OK with this change.

Roger-luo:
fyi. https://github.com/QuEraComputing/kirin/pull/370 is in v0.16.7 now

Roger-luo:
test failure seems due to a simple typo

src/bloqade/analysis/address/impls.py
david-pl: Changed it!

Roger-luo:
bump @weinbe58 for a review.

weinbe58 (approved):
LGTM

PR #165, opened 2025-04-16
Fix native cu3 rewrite if clause

david-pl:
Since #163 might be a while (depending on a change in kirin and whether we are fixing more pyright issues), I figured it might make sense to split out the bug fix. I cherry-picked the commit from the other PR.

The test isn't great, but it fails on the current main branch and works here, so better than nothing.


weinbe58 (approved):
LGTM!

PR #174, opened 2025-04-18
Support measuring registers in pyqrack

### LINKED ISSUE #160: PyQrack simulation backend fails to measure whole registers (only single qubit measurements work)
QASM allows measuring both single qubits and entire registers.  The QASM dialect and the simulation backend need to support both as well.  Multiple groups at YQuantum had to write code like this to measure the entire register:
```python
qasm2.measure(qreg[0], creg[0])
qasm2.measure(qreg[1], creg[1])
qasm2.measure(qreg[2], creg[2])
qasm2.measure(qreg[3], creg[3])
```

Example:
```python
from bloqade import qasm2
from bloqade.pyqrack import PyQrack
from collections import Counter

@qasm2.extended
def main():
    qreg = qasm2.qreg(4)
    creg = qasm2.creg(4)
    qasm2.measure(qreg, creg)  # Doesn't work in PyQrack
    #qasm2.measure(qreg[0], creg[0])  # Works
    return creg

device = PyQrack(dynamic_qubits=True, pyqrack_options={"isBinaryDecisionTree": False})
results = device.multi_run(main, _shots=100)
counts = Counter(map(lambda result:"".join(map(str, result)), results))
for key, value in counts.items():
    print(key, value)
```
Output:
```python
.../.venv/lib/python3.12/site-packages/cotengra/hyperoptimizers/hyper.py:54: UserWarning: Couldn't find `optuna`, `cmaes`, or `nevergrad` so will use completely random sampling in place of hyper-optimization.
  warnings.warn(
.../.venv/lib/python3.12/site-packages/cotengra/hyperoptimizers/hyper.py:36: UserWarning: Couldn't import `kahypar` - skipping from default hyper optimizer and using basic `labels` method instead.
  warnings.warn(
.../.venv/lib/python3.12/site-packages/cotengra/hyperoptimizers/hyper.py:73: UserWarning: Couldn't find `optuna`, `cmaes`, or `nevergrad` so will use completely random sampling in place of hyper-optimization.
  warnings.warn(

AttributeError                            Traceback (most recent call last)
Cell In[1], line 14
     11     return creg
     13 device = PyQrack(dynamic_qubits=True, pyqrack_options={"isBinaryDecisionTree": False})
---> 14 results = device.multi_run(main, _shots=100)
     15 counts = Counter(map(lambda result:"".join(map(str, result)), results))
     16 for key, value in counts.items():

File .../.venv/lib/python3.12/site-packages/bloqade/pyqrack/target.py:109, in PyQrack.multi_run(self, mt, _shots, *args, **kwargs)
    107 batched_results = []
    108 for _ in range(_shots):
--> 109     batched_results.append(interpreter.run(mt, args, kwargs).expect())
    111 return batched_results

File .../.venv/lib/python3.12/site-packages/kirin/interp/base.py:146, in BaseInterpreter.run(self, mt, args, kwargs)
    144 args = self.get_args(mt.arg_names[len(args) + 1 :], args, kwargs)
    145 try:
--> 146     _, results = self.run_method(mt, args)
    147 except InterpreterError as e:
    148     # NOTE: initialize will create new State
    149     # so we don't need to copy the frames.
    150     return Err(e, self.state.frames)

File .../.venv/lib/python3.12/site-packages/kirin/interp/concrete.py:30, in Interpreter.run_method(self, method, args)
     27 def run_method(
     28     self, method: Method, args: tuple[Any, ...]
     29 ) -> tuple[Frame[Any], Any]:
---> 30     return self.run_callable(method.code, (method,) + args)

File .../.venv/lib/python3.12/site-packages/kirin/interp/base.py:220, in BaseInterpreter.run_callable(self, code, args)
    218     return self.state.pop_frame(), self.void
    219 frame.set_values(body.blocks[0].args, args)
--> 220 results = self.run_callable_region(frame, code, body)
    221 return self.state.pop_frame(), results

File .../.venv/lib/python3.12/site-packages/kirin/interp/base.py:234, in BaseInterpreter.run_callable_region(self, frame, code, region)
    223 def run_callable_region(
    224     self, frame: FrameType, code: Statement, region: Region
    225 ) -> ValueType:
    226     """A hook defines how to run the callable region given
    227     the interpreter context. Frame should be pushed before calling
    228     this method and popped after calling this method.
   (...)    232     to be compatible with the Python convention.
    233     """
--> 234     results = self.run_ssacfg_region(frame, region)
    235     if isinstance(results, ReturnValue):
    236         return results.value

File .../.venv/lib/python3.12/site-packages/kirin/interp/concrete.py:37, in Interpreter.run_ssacfg_region(self, frame, region)
     35 block = region.blocks[0]
     36 while block is not None:
---> 37     results = self.run_block(frame, block)
     38     if isinstance(results, Successor):
     39         block = results.block

File .../.venv/lib/python3.12/site-packages/kirin/interp/concrete.py:55, in Interpreter.run_block(self, frame, block)
     53 frame.stmt = stmt
     54 frame.lino = stmt.source.lineno if stmt.source else 0
---> 55 stmt_results = self.eval_stmt(frame, stmt)
     56 if isinstance(stmt_results, tuple):
     57     frame.set_values(stmt._results, stmt_results)

File .../.venv/lib/python3.12/site-packages/kirin/interp/base.py:335, in BaseInterpreter.eval_stmt(self, frame, stmt)
    333 method = self.lookup_registry(frame, stmt)
    334 if method is not None:
--> 335     results = method(self, frame, stmt)
    336     if self.debug and not isinstance(results, (tuple, SpecialValue)):
    337         raise InterpreterError(
    338             f"method must return tuple or SpecialResult, got {results}"
    339         )

File .../.venv/lib/python3.12/site-packages/kirin/registry.py:19, in StatementImpl.__call__(self, interp, frame, stmt)
     18 def __call__(self, interp, frame, stmt: "Statement"):
---> 19     return self.impl(self.parent, interp, frame, stmt)

File .../.venv/lib/python3.12/site-packages/bloqade/pyqrack/qasm2/core.py:56, in PyQrackMethods.measure(self, interp, frame, stmt)
     54 qarg: PyQrackQubit = frame.get(stmt.qarg)
     55 carg: CBitRef = frame.get(stmt.carg)
---> 56 if qarg.is_active():
     57     carg.set_value(Measurement(qarg.sim_reg.m(qarg.addr)))
     58 else:

AttributeError: 'PyQrackReg' object has no attribute 'is_active'
```

david-pl:
Fixes #160

### DIFF
--- a/src/bloqade/qasm2/_wrappers.py
+++ b/src/bloqade/qasm2/_wrappers.py
@@ -73,19 +73,19 @@ def measure(qarg: Qubit, cbit: Bit) -> None:
 
 
 @overload
-def measure(qarg: QReg, cbit: CReg) -> None:
+def measure(qarg: QReg, carg: CReg) -> None:
     """
-    Measure the qubit `qarg` and store the result in the classical bit `cbit`.
+    Measure each qubit in the quantum register `qarg` and store the result in the classical register `carg`.
 
     Args:
-        qarg: The qubit to measure.
-        cbit: The classical bit to store the result in.
+        qarg: The quantum register to measure.
+        carg: The classical bit to store the result in.
     """
     ...
 
 
 @wraps(core.Measure)
-def measure(*args) -> None: ...
+def measure(qarg, carg) -> None: ...
 
 
 @wraps(uop.CX)
--- a/src/bloqade/qasm2/_wrappers.py
+++ b/src/bloqade/qasm2/_wrappers.py
@@ -59,7 +59,7 @@ def reset(qarg: Qubit) -> None:
 
 
 @wraps(core.Measure)
-def measure(qarg: Qubit, cbit: Bit) -> None:
+def measure(qarg: Qubit | QReg, cbit: Bit | CReg) -> None:
     """
     Measure the qubit `qarg` and store the result in the classical bit `cbit`.
 
--- a/src/bloqade/qasm2/_wrappers.py
+++ b/src/bloqade/qasm2/_wrappers.py
@@ -1,3 +1,5 @@
+from typing import overload
+
 from kirin.lowering import wraps
 
 from .types import Bit, CReg, QReg, Qubit
@@ -58,8 +60,8 @@ def reset(qarg: Qubit) -> None:
     ...
 
 
-@wraps(core.Measure)
-def measure(qarg: Qubit | QReg, cbit: Bit | CReg) -> None:
+@overload
+def measure(qarg: Qubit, cbit: Bit) -> None:
     """
     Measure the qubit `qarg` and store the result in the classical bit `cbit`.
 
@@ -68,6 +70,22 @@ def measure(qarg: Qubit | QReg, cbit: Bit | CReg) -> None:
         cbit: The classical bit to store the result in.
     """
     ...
+
+
+@overload
+def measure(qarg: QReg, cbit: CReg) -> None:
+    """
+    Measure the qubit `qarg` and store the result in the classical bit `cbit`.
+
+    Args:
+        qarg: The qubit to measure.
+        cbit: The classical bit to store the result in.
+    """
+    ...
+
+
+@wraps(core.Measure)
+def measure(*args) -> None: ...
 
 
 @wraps(uop.CX)
--- a/src/bloqade/qasm2/dialects/core/stmts.py
+++ b/src/bloqade/qasm2/dialects/core/stmts.py
@@ -46,9 +46,9 @@ class Measure(ir.Statement):
 
     name = "measure"
     traits = frozenset({lowering.FromPythonCall()})
-    qarg: ir.SSAValue = info.argument(QubitType)
+    qarg: ir.SSAValue = info.argument(QubitType | QRegType)
     """qarg (Qubit): The qubit to measure."""
-    carg: ir.SSAValue = info.argument(BitType)
+    carg: ir.SSAValue = info.argument(BitType | CRegType)
     """carg (Bit): The bit to store the result in."""
 
 
--- a/src/bloqade/qasm2/dialects/core/stmts.py
+++ b/src/bloqade/qasm2/dialects/core/stmts.py
@@ -47,9 +47,9 @@ class Measure(ir.Statement):
     name = "measure"
     traits = frozenset({lowering.FromPythonCall()})
     qarg: ir.SSAValue = info.argument(QubitType | QRegType)
-    """qarg (Qubit): The qubit to measure."""
+    """qarg (Qubit | QReg): The qubit or quantum register to measure."""
     carg: ir.SSAValue = info.argument(BitType | CRegType)
-    """carg (Bit): The bit to store the result in."""
+    """carg (Bit | CReg): The bit or register to store the result in."""
 
     def check_type(self) -> None:
         qarg_is_qubit = self.qarg.type.is_subseteq(QubitType)
--- a/src/bloqade/qasm2/dialects/core/stmts.py
+++ b/src/bloqade/qasm2/dialects/core/stmts.py
@@ -51,6 +51,17 @@ class Measure(ir.Statement):
     carg: ir.SSAValue = info.argument(BitType | CRegType)
     """carg (Bit): The bit to store the result in."""
 
+    def check_type(self) -> None:
+        qarg_is_qubit = self.qarg.type.is_subseteq(QubitType)
+        carg_is_bit = self.carg.type.is_subseteq(BitType)
+        if (qarg_is_qubit and not carg_is_bit) or (not qarg_is_qubit and carg_is_bit):
+            raise ir.TypeCheckError(
+                self,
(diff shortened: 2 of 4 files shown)


cduck:
Is there any type checking concern @Roger-luo with making instructions polymorphic like this?

src/bloqade/qasm2/dialects/core/stmts.py
```diff
@@ -46,9 +46,9 @@ class Measure(ir.Statement):
 
     name = "measure"
     traits = frozenset({lowering.FromPythonCall()})
-    qarg: ir.SSAValue = info.argument(QubitType)
+    qarg: ir.SSAValue = info.argument(QubitType | QRegType)
     """qarg (Qubit): The qubit to measure."""
-    carg: ir.SSAValue = info.argument(BitType)
+    carg: ir.SSAValue = info.argument(BitType | CRegType)
```
Roger-luo, line 51: you need to add a `def check_type(self)` method to validate the type here because here your definition allows `QubitType` appear with `CRegType`.

src/bloqade/qasm2/_wrappers.py
```diff
@@ -59,7 +59,7 @@ def reset(qarg: Qubit) -> None:
 
 
 @wraps(core.Measure)
-def measure(qarg: Qubit, cbit: Bit) -> None:
+def measure(qarg: Qubit | QReg, cbit: Bit | CReg) -> None:
```
Roger-luo, line 62: so you can fix the type hinting by adding the following separately

```python
from typing import overload

@overload
def measure(qarg: Qubit, cbit: Bit) -> None: ...

@overload
def measure(qarg: QReg, cbit: CReg) -> None: ...
```

Roger-luo (changes requested):
@cduck yes, the type hinting here is not correct. I've commented with a fix above.

src/bloqade/qasm2/_wrappers.py
Roger-luo: this is what it looks like in IDE

image

david-pl: @Roger-luo I had already done that following your message on slack. I just forgot to push the commit yesterday :sweat_smile:

src/bloqade/qasm2/dialects/core/stmts.py
david-pl: I added the type check test. FWIW, mixing the types would have already triggered a `RuntimeError` in the impl.

david-pl:
@Roger-luo CI now fails because I added a test for the `TypeCheckError` that is now raised. However, as of kirin v0.16.8, the resulting error is actually the `NoPythonStacktrace` base class.

So essentially, this doesn't work as you'd expect:
```python
try:
    raise ir.TypeCheckError(...)
except ir.TypeCheckError:
    ... # this is never called
except kirin.exception.NoPythonStacktrace:
   ... # <--- you end up here
```
Should this be considered a bug in kirin? There are a bunch of different error types that inherit from `NoPythonStacktrace`, so catching that may catch other errors too.

src/bloqade/qasm2/dialects/core/stmts.py
weinbe58: It would be better to have the error at compile time compared to run time since in some cases you might not run the function with the simulator for awhile if at all.

src/bloqade/qasm2/dialects/core/stmts.py
```diff
@@ -46,11 +46,22 @@ class Measure(ir.Statement):
 
     name = "measure"
     traits = frozenset({lowering.FromPythonCall()})
-    qarg: ir.SSAValue = info.argument(QubitType)
+    qarg: ir.SSAValue = info.argument(QubitType | QRegType)
     """qarg (Qubit): The qubit to measure."""
```
weinbe58, line 50: Don't forget an out the doc strings!

weinbe58 (commented):
LGTM once the CI is fixed

Roger-luo:
right I didn't realize that was actually breaking. I'm cleaning this up in https://github.com/QuEraComputing/kirin/pull/376 so it throws and validation error. We don't have a good way to raise a native exception while hiding python stacktrace at the moment because we we implementing this as custom error handler.

Roger-luo:
I figured out a way to raise the exact error that you raise inside the interpreter - it might a bit evil but I think this is cleaner on user side. stay tuned for a new kirin release.

Roger-luo:
but I think we should merge this PR given this is due to 0.16.8 actually breaking the convention a bit here. Can you fix the test based on https://github.com/QuEraComputing/kirin/pull/376 first so we can merge this PR but we can cleanup the exception type a bit later?

weinbe58:
Mark the test as xfail

david-pl:
> Mark the test as xfail
@weinbe58 the fix from @Roger-luo is already released in v0.16.9, so I guess that comment is no longer relevant?

PR #200, opened 2025-04-24
New Squin statements. 

weinbe58:
This PR is addressing the new squin statements. However, after discussion with @kaihsin and @Roger-luo we have slightly modified the semantics to make things a bit cleaner:

1. `Boradcast` is now an appliy-like statement, see #206 
2.  `PualiString` is now `CliffordString` to enable better rewrites to STIM.

### DIFF
--- a/src/bloqade/squin/op/stmts.py
+++ b/src/bloqade/squin/op/stmts.py
@@ -22,17 +22,6 @@ class CompositeOp(Operator):
     pass
 
 
-@statement(dialect=dialect)
-class Broadcast(CompositeOp):
-    """This takes an operator and allows it to be applied to a list of qubits."""
-
-    traits = frozenset(
-        {ir.Pure(), lowering.FromPythonCall(), MaybeUnitary(), HasSites()}
-    )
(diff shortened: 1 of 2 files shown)


johnzl-777:
@weinbe58 is there a reason we don't want broadcast in the wire dialect as well? Or is the idea that there will be some lowering from a qubit broadcast to a bunch of wire applications?

src/bloqade/squin/op/stmts.py
```diff
@@ -138,6 +146,18 @@ class PauliOp(ConstantUnitary):
     pass
 
 
+@statement(dialect=dialect)
+class CliffordString(ConstantUnitary):
```
cduck, line 150: What is a CliffordString supposed to represent?  PauliString has specific uses like defining a multi-qubit operator for measurement, control, or phase rotation.  I am not familiar with "Clifford string".

kaihsin: Clifford include also CX, do we also want product of mixing 2Q and 1Q?

PR #207, opened 2025-04-25
Implement pyqrack interpreter methods for squin dialect

### LINKED ISSUE #185: PyQrack support for squin
supporting a direct interpretation of squin dialect, so we can run simulation via squin directly - while we support QASM2 I think it makes sense to have both but primarily support squin because squin is more expressive than QASM2.

david-pl:
Closes #185


david-pl:
So, the implementation here is somewhat complete. However, I'm still struggling with the wire dialect.

@weinbe58 @johnzl-777 how do you even create a `Wire` right now? I tried this:
```python
@squin.wired
def main():
    q = squin.qubit.new(1)
    w = squin.wire.unwrap(q[0])
    x = squin.op.x()
    squin.wire.apply(x, w)
    return w
```
But, of course, `squin.qubit.new` isn't supported by the `wired` dialect. Is there something like a `NewWire` statement that is missing?

src/bloqade/pyqrack/squin/qubit.py
```diff
@@ -0,0 +1,80 @@
...
+    def measure(
+        self, interp: PyQrackInterpreter, frame: interp.Frame, stmt: qubit.Measure
+    ):
+        qubits: ilist.IList[PyQrackQubit, Any] = frame.get(stmt.qubits)
+        result = []
+        for qbit in qubits:
+            if qbit.is_active():
+                result.append(qbit.sim_reg.m(qbit.addr))
+            else:
+                result.append(None)
```
david-pl, line 55: @weinbe58 not sure about this one. We don't have a concept of a classical register in squin.

david-pl:
@weinbe58 two more things:

* I'm not sure about the `impl` for `MeasureAny`, please have a look.
* I could also simplify the `OperatorRuntime` by simply assuming that whenever `len(qubits) > 1` we are looking at a controlled gate. That would mean I could remove the `control_apply` method from all runtime classes, but I'm not sure how safe that assumption is. What do you think?

weinbe58:
@david-pl I can handle the measurement stuff later on thanks you!

src/bloqade/pyqrack/squin/qubit.py
```diff
@@ -0,0 +1,111 @@
...
+    @interp.impl(qubit.Apply)
+    def apply(self, interp: PyQrackInterpreter, frame: interp.Frame, stmt: qubit.Apply):
+        qubits: ilist.IList[PyQrackQubit, Any] = frame.get(stmt.qubits)
+
+        for qbit in qubits:
+            if not qbit.is_active():
+                return ()
+
+        operator: OperatorRuntimeABC = frame.get(stmt.operator)
+        operator.apply(*qubits)
```
weinbe58, line 36:
```suggestion
        
        active_qubits = [qubit for qubit in qubits if qubit.is_active()]
        operator: OperatorRuntimeABC = frame.get(stmt.operator)
        operator.apply(*active_qubits)
```

weinbe58: You should not exit the method just because an atom is lost.

src/bloqade/pyqrack/squin/runtime.py
```diff
@@ -0,0 +1,323 @@
...
+
+    def control_apply(self, *qubits: PyQrackQubit, adjoint: bool = False) -> None:
+        raise RuntimeError(f"Can't apply controlled version of {self}")
+
+    def broadcast_apply(self, qubits: ilist.IList[PyQrackQubit, Any], **kwargs) -> None:
+        for qbit in qubits:
+            if not qbit.is_active():
+                continue
+
+            self.apply(qbit, **kwargs)
```
weinbe58, line 26: What if the operator accepts two qubits? you should batch the qubits based on the operator size.

src/bloqade/pyqrack/squin/runtime.py
```diff
@@ -0,0 +1,323 @@
...
+            method_name += "mc"
+
+        if adjoint and self.method_name in ("s", "t"):
+            method_name += "adj"
+
+        return method_name + self.method_name
+
+    def apply(self, *qubits: PyQrackQubit, adjoint: bool = False) -> None:
+        method_name = self.get_method_name(adjoint=adjoint, control=False)
+        getattr(qubits[0].sim_reg, method_name)(qubits[0].addr)
```
weinbe58, line 53:
```suggestion
    def apply(self, qubits: PyQrackQubit, adjoint: bool = False) -> None:
        if not qubit.is_active():
            return
        method_name = self.get_method_name(adjoint=adjoint, control=False)
        getattr(qubits.sim_reg, method_name)(qubits.addr)
```

weinbe58: I assume If you expect to only have one qubit you don't need to have the `*qubits` here.

src/bloqade/pyqrack/squin/runtime.py
```diff
@@ -0,0 +1,323 @@
...
+
+    def control_apply(self, *qubits: PyQrackQubit, adjoint: bool = False) -> None:
+        ctrls = [qbit.addr for qbit in qubits[:-1]]
+        target = qubits[-1]
+        method_name = self.get_method_name(adjoint=adjoint, control=True)
+        getattr(target.sim_reg, method_name)(ctrls, target.addr)
```
weinbe58, line 59:
```suggestion
        ctrls = [qbit.addr for qbit in qubits[:-1], if qbit.is_active()]
        if len(ctrls) == 0: # do not apply if there are no control qubits. 
            return
        target = qubits[-1]
        method_name = self.get_method_name(adjoint=adjoint, control=True)
        getattr(target.sim_reg, method_name)(ctrls, target.addr)
```

src/bloqade/pyqrack/squin/runtime.py
```diff
@@ -0,0 +1,323 @@
...
+    def apply(self, *qubits: PyQrackQubit, adjoint: bool = False) -> None:
+        self.op.apply(*qubits, adjoint=adjoint)
+
+        target = qubits[-1]
+
+        # NOTE: just factor * eye(2)
+        m = self.mat(adjoint)
+
+        # TODO: output seems to always be normalized -- no-op?
+        target.sim_reg.mtrx(m, target.addr)
```
weinbe58, line 164: if this is true we might want to error instead of allowing things to run. If we want to do something like `rot(scale * op)` we can rewrite this to a standard rotation statement. 

cc: @Roger-luo

src/bloqade/pyqrack/squin/runtime.py
```diff
@@ -0,0 +1,323 @@
...
+            raise RuntimeError(
+                f"Rotation only supported for Pauli operators! Got {self.axis}"
+            )
+
+        try:
+            axis = getattr(Pauli, "Pauli" + self.axis.method_name.upper())
+        except KeyError:
+            raise RuntimeError(
+                f"Rotation only supported for Pauli operators! Got {self.axis}"
+            )
```
weinbe58, line 249: Since this is a runtime specific to PyQrack you can put this check as a validation inside the `__post_init__` of this class and save the result as `pyqrack_axis` field

src/bloqade/pyqrack/squin/runtime.py
```diff
@@ -0,0 +1,323 @@
...
+            return self.theta, self.phi, self.lam
+
+    def apply(self, *qubits: PyQrackQubit, adjoint: bool = False) -> None:
+        target = qubits[-1]
+        angles = self.angles(adjoint=adjoint)
+        target.sim_reg.u(target.addr, *angles)
+
+    def control_apply(self, *qubits: PyQrackQubit, adjoint: bool = False) -> None:
+        target = qubits[-1]
+        ctrls = [qbit.addr for qbit in qubits[:-1]]
```
weinbe58, line 306: same as comment above.

src/bloqade/squin/op/__init__.py
```diff
@@ -8,19 +8,27 @@
...
 
 
 @_wraps(stmts.Adjoint)
-def adjoint(op: types.Op, *, is_unitary: bool = False) -> types.Op: ...
+def adjoint(op: types.Op) -> types.Op: ...
 
 
 @_wraps(stmts.Control)
-def control(op: types.Op, *, n_controls: int, is_unitary: bool = False) -> types.Op: ...
+def control(op: types.Op, *, n_controls: int) -> types.Op: ...
```
weinbe58, line 27: Remove wrappers later when syntax sugar is implemented.

src/bloqade/pyqrack/squin/qubit.py
```diff
@@ -0,0 +1,111 @@
...
+        )
+        return (qreg,)
+
+    @interp.impl(qubit.Apply)
+    def apply(self, interp: PyQrackInterpreter, frame: interp.Frame, stmt: qubit.Apply):
+        qubits: ilist.IList[PyQrackQubit, Any] = frame.get(stmt.qubits)
+
+        for qbit in qubits:
+            if not qbit.is_active():
+                return ()
```
weinbe58, line 33: Let's leave this logic to the `operator.apply`

src/bloqade/pyqrack/squin/qubit.py
```diff
@@ -0,0 +1,111 @@
...
+
+    @interp.impl(qubit.MeasureQubitList)
+    def measure_qubit_list(
+        self,
+        interp: PyQrackInterpreter,
+        frame: interp.Frame,
+        stmt: qubit.MeasureQubitList,
+    ):
+        qubits: ilist.IList[PyQrackQubit, Any] = frame.get(stmt.qubits)
+        result = [self._measure_qubit(qbit) for qbit in qubits]
```
weinbe58, line 58: this should be an `ilist.IList` not a regular list. The hint in the wrapper is wrong.

src/bloqade/pyqrack/squin/qubit.py
```diff
@@ -0,0 +1,111 @@
...
+    def broadcast(
+        self, interp: PyQrackInterpreter, frame: interp.Frame, stmt: qubit.Broadcast
+    ):
+        operator: OperatorRuntimeABC = frame.get(stmt.operator)
+        qubits: ilist.IList[PyQrackQubit, Any] = frame.get(stmt.qubits)
+        operator.broadcast_apply(qubits)
+
+    def _measure_qubit(self, qbit: PyQrackQubit):
+        if qbit.is_active():
+            return qbit.sim_reg.m(qbit.addr)
```
weinbe58, line 48:
```suggestion
            return bool(qbit.sim_reg.m(qbit.addr))
```
this returns 0, 1 as an integer.

src/bloqade/pyqrack/squin/wire.py
```diff
@@ -0,0 +1,69 @@
...
+        interp: PyQrackInterpreter,
+        frame: interp.Frame,
+        stmt: wire.MeasureAndReset,
+    ):
+        w: PyQrackWire = frame.get(stmt.wire)
+        qbit = w.qubit
+        res: int = qbit.sim_reg.m(qbit.addr)
+        qbit.sim_reg.force_m(qbit.addr, False)
+        new_w = PyQrackWire(qbit)
+        return (new_w, res)
```
weinbe58, line 61: I need to make sure this will work in general. I know using `force_m` can create some issues. Lets leave it for now but I will revisit how to properly reset a qubit with pyqrack.

weinbe58 (changes requested):
I have some suggestions mostly for the runtime.

src/bloqade/pyqrack/squin/qubit.py
david-pl: @weinbe58 I quit the method because any qubit in the tuple `qubits` is either a control or a target. I'm not sure what should happen if any of those is lost, right?
Also, this is kind of consistent with how it's done for qasm2, e.g. here: https://github.com/QuEraComputing/bloqade-circuit/blob/6a35f856c00568cb244a2a2b59c4512d6a835c18/src/bloqade/pyqrack/qasm2/uop.py#L144

Also:
```python
active_qubits = [qubit for qubit in qubits if qubit.is_active()]
```
Doesn't this potentially mess up the order of controls and/or the target?

src/bloqade/pyqrack/squin/runtime.py
david-pl: This will result in typing issues though. I couldn't figure out how to overload the base `apply` method to accept both, because both variants return `None`. Do you have an idea?

src/bloqade/pyqrack/squin/runtime.py
david-pl: @weinbe58 this still applies to the target even if `target.is_active() is False`. That's inconsistent with the `apply` method now though, right?

src/bloqade/pyqrack/squin/qubit.py
david-pl: > What physically happens if either the control or target atom is lost and you try to apply a CZ gate is this: nothing.

@cduck Yes, I agree.

> But CZ gates do still occur on the other (non-lost) pairs of atoms.

What atom pairs do you mean? The `qubits` we are considering here are the ones explicitly passed into the `apply`, e.g. `apply(CZ, q[0], q[1])`. It's fine to bail if any of those are inactive then, right?

Am I misunderstanding something here?

src/bloqade/pyqrack/squin/runtime.py
david-pl: Which comment are you referring to?

src/bloqade/pyqrack/squin/qubit.py
cduck: Question for @weinbe58: Do SQUIN gate operations apply to multiple sets of qubits?  E.g. is `apply(CZ, q0, q1, q2, q3)` valid and equivalent to `apply(CZ, q0, q1); apply(CZ, q2, q3)`?

If the answer is no, then bailing if either atom is missing should be correct.  If the answer is yes, then you can't bail, you need to remove the correct *pair* of qubits so that specific instance of the gate doesn't occur.

src/bloqade/pyqrack/squin/qubit.py
david-pl: @weinbe58 I thought about this. It's either checking once here, or in every `apply` method in the runtime. Are you sure we'd want that? Of course, I can also make it an abstract method that checks this and then calls e.g. `_apply`, but that's also not so pretty.

src/bloqade/pyqrack/squin/qubit.py
david-pl: @cduck My comment above was a bit off, actually. Right now, the syntax for controlled gates in squin is a bit odd. You write something like `cz = control(z, n_controls=1)`, or `ccz = control(z, n_controls=2)`. Then, you can do e.g. `apply(cz, q[0], q[1])`. However, you could also do `apply(cz, q[0], q[1], q[2])` (`cz` here is not a typo).
In the latter case, I'm actually not entirely sure what the semantics are. Right now I went for "apply a `cz` where the first `n_controls` qubits are the controls and the last one is the target". That would mean that we apply a `cz q[0], q[2]` in this example.

So you were right, there are qubits in that list here that don't have anything to do with the gate being applied and we shouldn't bail. That is, if these are the semantics we want to go for here.

src/bloqade/pyqrack/squin/qubit.py
david-pl: Never mind, as discussed below, we can have situations where there are qubits passed in that have nothing to do with the gate that is being applied and we should just ignore those.

src/bloqade/pyqrack/squin/runtime.py
cduck: If it's a Rydberg-interaction-type gate, then a missing atom acts like the |0> state.  Often this is equivalent to no gate occurring.

src/bloqade/pyqrack/squin/qubit.py
cduck: I don't understand the semantics of `apply(cz, q[0], q[1], q[2])`.  Does this mean CZ(q0, q1)???

david-pl: > I don't understand the semantics 

Yes :sweat_smile: this will be a source of confusion for a lot of people, and I'm not happy about it.

To clarify: right now it means `CZ(q0, q2)` as I always choose the last qubit as the target.

The reason behind this is how generic the API for defining controls is right now: you can define `Control(A, n_controls=n)` for any operator `A` and any `n`. At the point where I'm implementing the runtime, I'm just applying any arbitrary operator with any number of controls.

Right now I'm thinking it might be better to error if `n_controls != len(qubits) - 1)`. What do you think @cduck @weinbe58 ?

cduck: Definitely error.  That is very non-intuitive.

cduck: > Yes 😅 this will be a source of confusion for a lot of people, and I'm not happy about it.

Then let's change the design?

Roger-luo: > don't understand the semantics of apply(cz, q[0], q[1], q[2]). Does this mean CZ(q0, q1)???

That's invalid code.

> Question for @weinbe58: Do SQUIN gate operations apply to multiple sets of qubits? E.g. is apply(CZ, q0, q1, q2, q3) valid and equivalent to apply(CZ, q0, q1); apply(CZ, q2, q3)?

That's broadcast.

> Right now I'm thinking it might be better to error if n_controls != len(qubits) - 1). What do you think @cduck @weinbe58 ?

Why you fix the operator at single side? This is wrong. You should actually validate on the operator size?

src/bloqade/squin/op/__init__.py
david-pl: Just out of curiosity: why? What's the hurt in keeping the wrapper?

david-pl:
Since this PR is getting kind of long, I'm gonna summarize some of the discussions again here. For the most part, I think there was a bit of confusion as I originally only considered a single target for controlled gates and broadcast (which was wrong). That has been fixed.

Here's an overview of the semantics we discussed in some comments and how I chose to implement them for now:

* In a `Control` you can have arbitrarily many controls and targets.
* If any of the controls are inactive, no gate is applied to any target.
* If the number of controls + the size of the applied operator doesn't match the number of qubits, you get a `RuntimeError`.
* If any of the targets are inactive, they are skipped. The gate is still applied to the remaining, active targets.
* The number of targets must match the size (I called the property `n_qubits`) of the operator that is to be applied to the targets (not just for controls, obviously).
* You can broadcast operators of arbitrary size. It's applied to the list of qubits in batches that match the size of the operator from left to right.
* That means, you can also broadcast controls, where e.g. `CX` is then applied to a list of qubits, where each pair is treated as a control and target, respectively. I'm not a huge fan of this since I think it might still be confusing, but I see the value in it.

Pinging @cduck , @Roger-luo and @weinbe58 here. Do you all agree with this?

Also, ready for another round of review from my side!

src/bloqade/squin/op/__init__.py
Roger-luo: what syntax sugar are we proposing here for control?

Roger-luo:
Yes, I agree, I think these are good summarize of the semantics, worth putting them in the docstring of the wrappers later.

just to be pedantic here

> The number of targets must match the size (I called the property n_qubits) of the operator that is to be applied to the targets (not just for controls, obviously).

we should call it `n_sites` because it is not necessarily n_qubits for operators.

weinbe58 (changes requested):

weinbe58 (approved):
LGTM, if we run into any issues we can polish them later.

PR #208, opened 2025-04-25
Add pyqrack extas with platform dependent constraints. 

weinbe58:
I am also making pyqrack-cpu implementation a default dependency since it has no other dependencies it is pretty light


Roger-luo:
if we choose this to be the default simulator sure, I don't see why not.

weinbe58:
OK I'm gonna merge this then

PR #211, opened 2025-04-25
Adding noise statements to squin

johnzl-777:
Some initial friction points/anticipated changes I can immediately see in translating this to stim

- The original `PauliChannel` definition from @kaihsin can take a variable number of float arguments, but in the stim dialect there are explicit fields for each probability of a 1Q/2Q pauli operator being applied. My feeling is that this should be reflected on the squin side as just something like `1QPauliChannel` and `2QPauliChannel` with explicit fields. I don't think we'd lose any generality here along with the added benefit that it would be much harder for a user to misinput information (the default definition assumes you plug in either 3 or 15 values via varargs but it seems to easy to make a mistake here)
  - ~~If we DO want varargs I'm not quite sure how to nicely feed this to the kirin `@wraps` decorator~~
- From #200 , I see it could be possible to not have to provide an operator for `basis` and instead use a `CliffordString` (although I'd have to restrict CliffordString even further, to just the Pauli Operators. Would it be worth having something like a `PauliString`? Or is that silly?)

cc: @weinbe58 @Roger-luo @kaihsin

### DIFF
--- a/src/bloqade/squin/noise/__init__.py
+++ b/src/bloqade/squin/noise/__init__.py
@@ -1,6 +1,5 @@
 # Put all the proper wrappers here
 
-# from kirin.types import Float, Tuple, Vararg
 from kirin.lowering import wraps as _wraps
 
 from bloqade.squin.op.types import Op
@@ -20,11 +19,8 @@ def pp_error(op: Op, p: float) -> Op: ...
 def depolarize(n_qubits: int, p: float) -> Op: ...
 
 
-# How do you give `types.Tuple[types.Vararg(types.Float)])` nicely to
-# the wrapper?
-
-# @_wraps(stmts.PauliChannel)
-# def pauli_channel(n_qubits: int, params:) -> Op:...
+@_wraps(stmts.PauliChannel)
+def pauli_channel(n_qubits: int, params: tuple[float, ...]) -> Op: ...
 
 
 @_wraps(stmts.QubitLoss)
--- a/src/bloqade/squin/noise/__init__.py
+++ b/src/bloqade/squin/noise/__init__.py
@@ -1,0 +1,31 @@
+# Put all the proper wrappers here
+
+# from kirin.types import Float, Tuple, Vararg
+from kirin.lowering import wraps as _wraps
+
+from bloqade.squin.op.types import Op
+
+from . import stmts as stmts
+
+
+@_wraps(stmts.PauliError)
+def pauli_error(basis: Op, p: float) -> Op: ...
+
+
+@_wraps(stmts.PPError)
+def pp_error(op: Op, p: float) -> Op: ...
+
+
+@_wraps(stmts.Depolarize)
+def depolarize(n_qubits: int, p: float) -> Op: ...
+
+
+# How do you give `types.Tuple[types.Vararg(types.Float)])` nicely to
+# the wrapper?
+
+# @_wraps(stmts.PauliChannel)
+# def pauli_channel(n_qubits: int, params:) -> Op:...
+
+
+@_wraps(stmts.QubitLoss)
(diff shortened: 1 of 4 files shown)


src/bloqade/squin/noise/__init__.py
```diff
@@ -0,0 +1,31 @@
...
+@_wraps(stmts.PPError)
+def pp_error(op: Op, p: float) -> Op: ...
+
+
+@_wraps(stmts.Depolarize)
+def depolarize(n_qubits: int, p: float) -> Op: ...
+
+
+# How do you give `types.Tuple[types.Vararg(types.Float)])` nicely to
+# the wrapper?
```
Roger-luo, line 24: that's `tuple[float, ...]`

src/bloqade/squin/noise/stmts.py
```diff
@@ -0,0 +1,64 @@
...
+
+
+@statement
+class NoiseChannel(ir.Statement):
+    pass
+
+
+@statement(dialect=dialect)
+class PauliError(NoiseChannel):
+    name = "pauli_error"
```
Roger-luo, line 16: this is not necessary, `@statement` will generate the name to snakecase

src/bloqade/squin/noise/stmts.py
```diff
@@ -0,0 +1,64 @@
...
+    op: ir.SSAValue = info.argument(OpType)
+    p: ir.SSAValue = info.argument(types.Float)
+    result: ir.ResultValue = info.result(OpType)
+
+
+@statement(dialect=dialect)
+class Depolarize(NoiseChannel):
+    """
+    Apply n-qubit depolaize error to qubits
+    NOTE For Stim, this can only accept 1 or 2 qubits
```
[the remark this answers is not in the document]
cduck: THIS SUGGESTED CHANGE DELETES THE FOLLOWING LINE.

johnzl-777:
This is the simplest port of the noise statements that @kaihsin originally produced. 

I just have some observations/ideas that I'd appreciate getting feedback on (if it's worth it, I can make an RFC)

- `PauliChannel` in this PR can take any number of floats but in Stim it boils down to either 3 values for a 1Q channel or 15 values for a 2Q channel. In the Stim dialect those numbers are fixed per statement. While I can certainly enforce the correct number of values during code generation I don't think it would hurt to just mirror that structure in squin.
- Should `PPError` in squin map to the CORRELATED_ERROR...ELSE_CORRELATED_ERROR functionality in Stim? In which case, that should be added into the Stim dialect itself?

johnzl-777:
Oh my goodness Copilot please go away I beg you

Roger-luo (approved):
LGTM can we have some basic tests on the constructors? Or maybe you could help @david-pl add the corresponding runtime in a new PR (which should test this too). I think would be nice to have it before next week's hackthon.

weinbe58:
@Roger-luo I think we could simplify the runtime if we added a target which would be able to handle a wider variety of channels and would be easy to implement with `PyQrack` using the current runtime in #185 

```python

@statement(dialect=dialect)
class StochasticUnitaryChannel
    operators: ir.SSAValue = argument(ilist.IListType[OpType, NumOperators]
    probabilities: ir.SSAValue = argument(ilist.IListType[float, NumOperators]

```

you could basically just rewrite all these other specialized noise channels to this statement and then the runtime would be just randomly select which operator to apply.

johnzl-777:
From discussing with @Roger-luo and @weinbe58 I agree this PR is good to go (apologies again for sitting on this!) and just wanted to put down two notes:
- @Roger-luo asked for some simple unit tests on the constructors, I'll make a separate PR for that
- @weinbe58 's proposed additional statement will also go into a separate PR 👍

PR #218, opened 2025-04-29
Add a simple circuit fidelity analysis pass

david-pl:


weinbe58:
@david-pl The issue is that you are trying to use the `qasm2.main` dialect with the noise which isn't supported because we need `ilist`

src/bloqade/analysis/fidelity/impls.py
```diff
@@ -0,0 +1,100 @@
...
+        interp._current_atom_survival_probability = 1.0
+        for s in stmt.body.stmts():
+            stmt_impl = interp.lookup_registry(frame=frame, stmt=s)
+            if stmt_impl is None:
+                continue
+
+            stmt_impl(interp=interp, frame=frame, stmt=s)
+
+        loop_body_gate_fidelity = interp._current_gate_fidelity
+        loop_body_atom_survival = interp._current_atom_survival_probability
```
weinbe58, line 90: logically this makes sense but I think you're reinventing the wheel in terms of executing the loop body.Take a look at how kirin implements the interpreters for the structural control flow: https://github.com/QuEraComputing/kirin/blob/2bd47e78eee36a3c703a01655e1af0f5e95492e0/src/kirin/dialects/scf/interp.py#L24
This can also be used for the abstract interpreter.

src/bloqade/noise/fidelity.py
```diff
@@ -0,0 +1,43 @@
...
+        interp._current_gate_fidelity *= fid
+
+    @interp.impl(AtomLossChannel)
+    def atom_loss(
+        self,
+        interp: FidelityAnalysis,
+        frame: interp.Frame[EmptyLattice],
+        stmt: AtomLossChannel,
+    ):
+        interp._current_atom_survival_probability *= 1 - stmt.prob
```
weinbe58, line 43: Would it be difficult to split this up to track the loss probability of individual qubits?

weinbe58 (changes requested):

src/bloqade/noise/fidelity.py
david-pl: Not sure. How would you keep track of which qubit is which? It would work for `PyQrackQubit`s because I can use the `addr`, but this is more general, and I just have an `IList` of qubits.

So you could have something like this:
```python
AtomLossChannel(prob=0.001, qargs=[q[0], q[1]])
AtomLossChannel(prob=0.001, qargs=[q[0], q[2]])
```
Do you have an idea how I could tell that we have `q[2]` in the second entry in the second line?

src/bloqade/analysis/fidelity/impls.py
```diff
@@ -0,0 +1,77 @@
...
+        self, interp: FidelityAnalysis, frame: interp.Frame[EmptyLattice], stmt: For
+    ):
+        if not isinstance(hint := stmt.iterable.hints.get("const"), const.Value):
+            # NOTE: not clear how long this loop is
+            # TODO: should we at least count the body once?
+            return
+
+        for _ in hint.data:
+            for s in stmt.body.stmts():
+                interp.eval_stmt(frame=frame, stmt=s)
```
david-pl, line 77: @weinbe58 like this?

src/bloqade/analysis/fidelity/impls.py
```diff
@@ -0,0 +1,77 @@
...
+from kirin.lattice import EmptyLattice
+from kirin.analysis import const
+from kirin.dialects import scf
+from kirin.dialects.scf.stmts import For, Yield, IfElse
+
+from .analysis import FidelityAnalysis
+
+
+@scf.dialect.register(key="circuit.fidelity")
+class ScfFidelityMethodTable(interp.MethodTable):
```
weinbe58, line 11: For this PR since we will be breaking the interpreter for regions in the next version of kirin can we remove the SCF interpretation here?

david-pl: As in, remove the entire method table? Or what do you mean?

weinbe58: Yeah we don't need to worry about control flow for this PR

src/bloqade/noise/fidelity.py
weinbe58: There are two options I guess: 

1. Create a new lattice and have a lattice element that accounts for the qubit values
2. Reuse the AddressAnalysis results and pass that into this interpreter and use those values inside the interpreter to map the qubit ssa values to qubit ids.

weinbe58 (approved):
LGTM

PR #224, opened 2025-05-01
Change Clifford String to Pauli String

johnzl-777:
Per yesterday's meeting Scientific Software meeting agreement, the Clifford string operator should be restricted to just Pauli's. 

@weinbe58 @Roger-luo I know there were several methods mentioned to translating the string into an actual `OpType` that can be used in other statements, ranging from lowering, canonicalization rewrite, etc. What is the _exact_ approach that should be taken for this? I'd like to open an issue to keep track of the infrastructure required for this.

cc: @cduck @kaihsin

### DIFF
--- a/src/bloqade/squin/op/stmts.py
+++ b/src/bloqade/squin/op/stmts.py
@@ -147,14 +147,14 @@ class PauliOp(ConstantUnitary):
 
 
 @statement(dialect=dialect)
-class CliffordString(ConstantUnitary):
+class PauliString(ConstantUnitary):
     traits = frozenset({ir.Pure(), lowering.FromPythonCall(), Unitary(), HasSites()})
     string: str = info.attribute()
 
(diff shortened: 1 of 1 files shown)


weinbe58 (approved):

weinbe58:
> I know there were several methods mentioned to translating the string into an actual OpType that can be used in other statements, ranging from lowering, canonicalization rewrite, etc. What is the exact approach that should be taken for this? I'd like to open an issue to keep track of the infrastructure required for this.


Correct me if I am wrong @Roger-luo but I thought the whole idea of having this was so that we could have a specific runtime that would not be dependent on implementing the runtime of `Kron` which is a bit non-trivial at the moment with pyqrack.

Roger-luo:
Yes it's like that we need to decide how much we want to support the generic Kron too.

PR #225, opened 2025-05-01
Update `PyQrack.run` to match RFC

weinbe58:
Closes https://github.com/QuEraComputing/bloqade/issues/220

### DIFF
--- a/src/bloqade/pyqrack/target.py
+++ b/src/bloqade/pyqrack/target.py
@@ -1,4 +1,4 @@
-from typing import List, TypeVar, ParamSpec
+from typing import Any, TypeVar, Iterator
 from dataclasses import field, dataclass
 
 from kirin import ir
@@ -12,9 +12,6 @@ from bloqade.pyqrack.base import (
     _default_pyqrack_args,
 )
 from bloqade.analysis.address import AnyAddress, AddressAnalysis
-
-Params = ParamSpec("Params")
-RetType = TypeVar("RetType")
 
 
 @dataclass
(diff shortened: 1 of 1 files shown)


src/bloqade/pyqrack/target.py
```diff
@@ -64,49 +63,50 @@ def _get_interp(self, mt: ir.Method[Params, RetType]):
 
     def run(
         self,
-        mt: ir.Method[Params, RetType],
-        *args: Params.args,
-        **kwargs: Params.kwargs,
-    ) -> RetType:
+        mt: ir.Method[..., RetType],
+        shots: int = 1,
```
Roger-luo, line 67: should `shots` be a keyword only?

src/bloqade/pyqrack/target.py
```diff
@@ -64,49 +63,50 @@ def _get_interp(self, mt: ir.Method[Params, RetType]):
...
+        args: tuple[Any, ...] = (),
+        kwargs: dict[str, Any] = {},
+        return_iterator: bool = False,
```
Roger-luo, line 70: what's the use case of `return_iterator`?

src/bloqade/pyqrack/target.py
```diff
@@ -64,49 +63,50 @@ def _get_interp(self, mt: ir.Method[Params, RetType]):
...
-            _shots (int):
-                The number of times to run the kernel method.
-
-        Returns
-            List of results of the kernel method, one for each shot.
+            RetType | list[RetType] | Iterator[RetType]:
+                The result of the simulation. If `return_iterator` is True,
+                an iterator that yields results for each shot is returned.
+                Otherwise, a list of results is returned if `shots > 1`, or
+                a single result is returned if `shots == 1`.
```
Roger-luo, line 95: I'm not sure if returning different data type here is a good idea. should we always return `list` or `Iterator[RetType]`? Maybe `Iterable[RetType]`?

weinbe58: This is a typo It should be `Iterable`

weinbe58: One option is we just always return the `Iterable` and then just let people capture those values into a list by default:

`result = list(device.run(...))`

johnzl-777 (approved):

PR #228, opened 2025-05-04
Task Interface

weinbe58:


Roger-luo (approved):
I would like to get this merged today so we can have a release. But I think there are few details we need to keep iterate on:

1. how we handle the return value, list[Future] vs. a single `Future` I think this needs to match how API works (as an implementation detail)
2. should we use `.run(kernel, *args, **kwargs)` or `.run(kernel, args: tuple, kwargs: dict`? I feel the former is more intuitive but I see it is necessary to distinguish the options in the task constructor. so I'm wondering if there are better ways?

src/bloqade/pyqrack/device.py
```diff
@@ -0,0 +1,166 @@
...
+from bloqade.analysis.address.analysis import AddressAnalysis
+
+RetType = TypeVar("RetType")
+Params = ParamSpec("Params")
+
+
+@dataclass
+class PyQrackSimulatorBase(AbstractSimulatorDevice[PyQrackSimulatorTask]):
+    options: PyQrackOptions = field(default_factory=_default_pyqrack_args)
+    loss_m_result: Measurement = field(default=Measurement.One, kw_only=True)
```
Roger-luo, line 29: I think this part will need some co-design with a target in Kirin. The pattern of running something on a device and wrap it in a task object is probably something we can generalize in Kirin.

weinbe58:
> I would like to get this merged today so we can have a release. But I think there are few details we need to keep iterate on:
> 
> 1. how we handle the return value, list[Future] vs. a single `Future` I think this needs to match how API works (as an implementation detail)

lets continue discussion in the RFC. 

> 2. should we use `.run(kernel, *args, **kwargs)` or `.run(kernel, args: tuple, kwargs: dict`? I feel the former is more intuitive but I see it is necessary to distinguish the options in the task constructor. so I'm wondering if there are better ways?

Actually, the former works quite well but the only issue is you can't have extra arguments for these methods because they will potentially clash with the ParamSpec. I tried to see if you could hint `args: tuple` with the ParamSpec but I don't think its possible. Ideally we could have support for `device.task(kernel: Method[Params, RetType], *args: Param.args, **kwargs: Params.kwargs) -> Task` and hide the optional arguments either when constructing the device or callint `task.run(**options)` but that would mean something like `device.run` wouldn't work with the type hints.

PR #233, opened 2025-05-06
Fix resetting qubits in pyqrack runtime

david-pl:


src/bloqade/pyqrack/squin/qubit.py
```diff
@@ -71,15 +71,23 @@ def measure_and_reset(
         result = []
         for qbit in qubits:
             if qbit.is_active():
-                result.append(qbit.sim_reg.m(qbit.addr))
+                qbit_result = bool(qbit.sim_reg.m(qbit.addr))
             else:
-                result.append(None)
-            qbit.sim_reg.force_m(qbit.addr, 0)
+                qbit_result = None
```
weinbe58, line 76: The Interpreter has a value `interp.loss_m_result` which stores the value of the measurement of a lost qubit.

src/bloqade/pyqrack/squin/wire.py
```diff
@@ -55,15 +55,22 @@ def measure_and_reset(
...
         w: PyQrackWire = frame.get(stmt.wire)
         qbit = w.qubit
-        res: int = qbit.sim_reg.m(qbit.addr)
-        qbit.sim_reg.force_m(qbit.addr, False)
+        res: bool = bool(qbit.sim_reg.m(qbit.addr))
+
+        if res:
+            qbit.sim_reg.x(qbit.addr)
+
+        # TODO: do we need to rewrap this here? The qbit changed in-place
```
weinbe58, line 63: I think there is a missing case checking if the qubit is active here.

weinbe58 (changes requested):
LGTM! just some minor changes related to atom loss stuff

src/bloqade/pyqrack/squin/qubit.py
weinbe58: I think we should update the `Measure` statement to return this enum instead of a bool actually this would be consistent with my discussions with @cduck

weinbe58: I think we should split that into another PR though

weinbe58 (approved):

Issue #238, opened 2025-05-06
QASM2.loads should have a default circuit name


Roger-luo:
I think `QASM2(dialects).loads(string, kernel_name)` is too complicated, we should re-export a `loads` in qasm2 module e.g `qasm2.loads` (hide the `qasm2.parse.loads` in `parse`) as following

```python
qasm2.loads(string) # dialects defaults to `qasm2.main`
```

similarly there should be a `loadfile`. The `name` can be generated via a UUID or a more readable random string.

PR #240, opened 2025-05-06
simplify loads

Roger-luo:
address #238 one question tho, was there a particular reason why the `loads` and `loadfile` was living inside the lowering target? @weinbe58 I think if not would be nice to move them out. This is not an object we expect users to use directly cuz it would be too confusing without knowing about lowering process.

### DIFF
--- a/src/bloqade/qasm2/__init__.py
+++ b/src/bloqade/qasm2/__init__.py
@@ -17,3 +17,4 @@ from .types import (
 )
 from .groups import gate as gate, main as main, extended as extended
 from ._wrappers import *  # noqa: F403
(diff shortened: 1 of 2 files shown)


weinbe58:
> address #238 one question tho, was there a particular reason why the `loads` and `loadfile` was living inside the lowering target? @weinbe58 I think if not would be nice to move them out. This is not an object we expect users to use directly cuz it would be too confusing without knowing about lowering process.

see the discussion: https://github.com/QuEraComputing/bloqade-circuit/pull/159

weinbe58 (approved):
LGTM

PR #250, opened 2025-05-07
Restrict qasm2 parser so it rejects multiple expressions in if body

david-pl:
Not sure if this is actually the intended solution, but at least it works.

Funny enough, I had to adjust the `main.qasm` test file since it included an invalid if statement. So we were actually testing that the invalid syntax worked before.


weinbe58:
gonna cc: @Roger-luo here. I Feel like this isn't the correct solution we can raise a better error inside the lowering pass which could give more specific source information.

Roger-luo:
well I tweaked the QASM2 a little bit to make it less restrictive cuz it was quite inconvenient but I see the issue here when using it as an interop format this becomes quite painful. So I think just removing this syntax is the right way to go.

PR #248, opened 2025-05-07
Raise error on invalid qasm2 if statement then body

david-pl:
@weinbe58 I'm not entirely sure about the `GateFunction`. Technically, I suppose I'd need to have more checks there.

Issue #155, opened 2025-04-15
QASM2 boolean operation on measurement results fails code-gen


cduck:
Bug report by https://github.com/randyshee/yquantum/blob/main/2025/team_solutions/charlie_and_delta/bug_report.ipynb

QASM2 supports a measurement operation followed by classical conditional logic on those measurement results.  Bloqade fails to lower this logic if there is a boolean operation as seen in the if-statement below.

```python
from bloqade import qasm2
from bloqade.qasm2.emit import QASM2 # the QASM2 target
from bloqade.qasm2.parse import pprint # the QASM2 pretty printer

@qasm2.extended
def circ_bool():
    qr = qasm2.qreg(4)
    cr = qasm2.creg(4)

    qasm2.measure(qr, cr)

    cond = cr[0] == 0 and cr[1] == 1  # Fails to lower boolean operation on measurement results
    if cond:
        qasm2.x(qr[3])

    # But this works instead
    #if cr[0] == 0:
    #    qasm2.creg(0)  # Dummy line to try to fix issue #157
    #    if cr[1] == 1:
    #        qasm2.x(qr[3])

    return qr


target = QASM2()
ast = target.emit(circ_bool)
pprint(ast)
```
Error:
```python
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[1], line 29
     25     return qr
     28 target = QASM2()
---> 29 ast = target.emit(circ_bool)
     30 pprint(ast)

File .../.venv/lib/python3.12/site-packages/bloqade/qasm2/emit/target.py:103, in QASM2.emit(self, entry)
    101 Py2QASM(entry.dialects)(entry)
    102 target_main = EmitQASM2Main(self.main_target)
--> 103 target_main.run(
    104     entry, tuple(ast.Name(name) for name in entry.arg_names[1:])
    105 ).expect()
    107 main_program = target_main.output
    108 assert main_program is not None, f"failed to emit {entry.sym_name}"

File .../.venv/lib/python3.12/site-packages/kirin/interp/base.py:146, in BaseInterpreter.run(self, mt, args, kwargs)
    144 args = self.get_args(mt.arg_names[len(args) + 1 :], args, kwargs)
    145 try:
--> 146     _, results = self.run_method(mt, args)
    147 except InterpreterError as e:
    148     # NOTE: initialize will create new State
    149     # so we don't need to copy the frames.
    150     return Err(e, self.state.frames)

[...]

PR #252, opened 2025-05-07
Implement pass that splits multiline ifs in qasm2

### LINKED ISSUE #189: Add rewrite pass QASM2Fold pass split if-statement
The idea here is to slit an if statement with multiple expressions inside of it to be in separate sequential if statements.

david-pl:
Closes #189 

Also, it will probably break the tests in #248


david-pl:
this also needs the check mentioned here: https://github.com/QuEraComputing/bloqade-circuit/pull/248#issuecomment-2858712156

david-pl:
The rewrite fails for nested if statements, but I'm not sure if we can handle them generally.

One such case, for example, is

```python
@qasm2.main
def main():
    q = qasm2.qreg(2)
    c = qasm2.creg(2)

    qasm2.h(q[0])
    qasm2.h(q[1])

    qasm2.measure(q, c)

    if c[0] == 1:
        if c[1] == 1:
            qasm2.x(q[0])

    return q
```

This would basically require boolean logic, since it can be rewritten to `c[0] == 1 and c[1] == 1`. There's another issue for that see #155, so I think it's fine to leave it for now.

@weinbe58 ready for review.

weinbe58:
> The rewrite fails for nested if statements, but I'm not sure if we can handle them generally.
> 
> One such case, for example, is
> 
> ```python
> @qasm2.main
> def main():
>     q = qasm2.qreg(2)
>     c = qasm2.creg(2)
> 
>     qasm2.h(q[0])
>     qasm2.h(q[1])
> 
>     qasm2.measure(q, c)
> 
>     if c[0] == 1:
>         if c[1] == 1:
>             qasm2.x(q[0])
> 
>     return q
> ```
> 
> This would basically require boolean logic, since it can be rewritten to `c[0] == 1 and c[1] == 1`. There's another issue for that see #155, so I think it's fine to leave it for now.
> 
> @weinbe58 ready for review.

I am still not sure if boolean operators are allowed in QASM2 so I think this shuold error.

test/qasm2/emit/test_qasm2_emit.py
```diff
@@ -258,9 +258,7 @@ def multiline_then():
         return q
 
     target = qasm2.emit.QASM2()
-
-    with pytest.raises(InterpreterError):
-        target.emit(multiline_then)
+    target.emit(multiline_then)
```
weinbe58, line 261: Maybe you can turn off the splitting here to get the error back?

weinbe58 (approved):
LGTM, I think you can probably restore the old test, see my comment. but Feel free to merge this

test/qasm2/emit/test_qasm2_emit.py
david-pl: Ah yes, I removed it before adding the kwarg. Will do!

Issue #255, opened 2025-05-08
UOpToParallel doesn't work with scf.


weinbe58:
```python
from bloqade import qasm2
from bloqade.qasm2.passes import UOpToParallel


@qasm2.extended
def main():
    q = qasm2.qreg(n_qubits=4)
    c = qasm2.creg(4)
    qasm2.h(q[0])
       
    qasm2.measure(q, c)

    if c[0] == 0:
        qasm2.x(q[0])
        qasm2.x(q[1])
    else:
        qasm2.x(q[2])
        qasm2.x(q[3])
        
    return q

UOpToParallel(main.dialects,rewrite_to_native_first=True)(main)
main.print()


```

This ideally would parallelize the if-blocks but currently it does not:

```
func.func main() -> !py.IList[!py.Qubit, !Any] {
  ^0(%main_self):
  │ %0 = py.constant.constant 4 : !py.int
  │ %q = qasm2.core.qreg.new(n_qubits=%0) : !py.IList[!py.Qubit, !Any]
  │ %c = qasm2.core.creg.new(n_bits=%0) : !py.CReg
  │ %1 = py.constant.constant 0 : !py.int
  │ %2 = qasm2.core.qreg.get(reg=%q, idx=%1) : !py.Qubit
  │ %3 = py.constant.constant 1.5707963267949 : !py.float
  │ %4 = py.constant.constant 0.0 : !py.float
  │ %5 = py.constant.constant 3.14159265358979 : !py.float
  │      qasm2.uop.U(qarg=%2, theta=%3, phi=%4, lam=%5)
  │      qasm2.core.measure(qarg=%q, carg=%c)
  │ %6 = qasm2.core.creg.get(reg=%c, idx=%1) : !py.Bit
  │ %7 = py.cmp.eq(lhs=%6, rhs=%1) : !py.bool
  │      scf.if %7 {
  │      │ ^1(%14):
  │      │ │  %8 = py.constant.constant 0 : !py.int
  │      │ │  %9 = qasm2.core.qreg.get(reg=%q, idx=%8) : !py.Qubit
  │      │ │ %10 = py.constant.constant 3.14159265358979 : !py.float
  │      │ │ %11 = py.constant.constant 0.0 : !py.float
  │      │ │       qasm2.uop.U(qarg=%9, theta=%10, phi=%11, lam=%10)
  │      │ │ %12 = py.constant.constant 1 : !py.int
[...]

PR #258, opened 2025-05-09
Fixing Measure and fixing issues with the qregister types

weinbe58:
Trying to fix some major problems:

1. Type inference wasn't inferring the length of the qubit register, this should be fixed now
2. Switching to using a rewrite pass to implement the syntax sugar for measure
3. Adding some fixes to `Py2QASM` and `QASM2Py` since the register is now an IList
4. Adding `HintLen` to QASM2Fold pass in order to help unroll loops which depend on `len` of a register

I've notice a few bugs in the current passes. namely #255 but also I'm trying to figure out an issue with `scf.PickIfElse` unroll pass which seems to be deleting code. trying to come up with a MWE

cc: @Roger-luo

### DIFF
--- a/src/bloqade/qasm2/_wrappers.py
+++ b/src/bloqade/qasm2/_wrappers.py
@@ -68,14 +68,20 @@ def measure(qreg: QReg, creg: CReg) -> None: ...
 def measure(qarg: Qubit, cbit: Bit) -> None: ...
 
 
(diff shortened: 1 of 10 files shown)


weinbe58:
I'm going to close this issue because I think we really need to refactor the dialects first before addressing these problems.

Issue #180, opened 2025-04-22
RFC: Facilitation of squin → stim translation with more flexible Reset/MeasureAndReset statements


johnzl-777:
This has some relation to the (nearly concluded?) Measurement RFC (#158 ) so I’ll do my best here to carry over what seems to be the consensus on that RFC over to what’s going on here. 

Basically, while Measure in squin has the issue of not quite translating well to Stim (or at least, requiring a bit more complexity to do so), MeasureAndReset/Reset seem more restrictive in their definitions than what the Stim dialect can support.

For example, In the ongoing squin → stim rewrite #148  (which is moving towards codegen btw) I assumed a `Reset` in squin maps to an `RZ` in the stim dialect and `MeasureAndReset` is an `MZ` followed by an `RZ` in Stim. However, it seems to me potentially beneficial (the QEC folks would have to weigh in on this) to allow a user to, just like in the Measurement RFC, define a basis for Reset:

```python
# Proposed modifications to the squin dialect
## in wire
class Reset:
	basis: ir.SSAValue = info.argument(OpType)
	wire: ir.SSAValue = info.argument(WireType)
	

## in qubit
class Reset:
	basis: ir.SSAValue = info.argument(OpType)
	qubits: ir.SSAValue = info.argument(IListType[Qubit, ...])
```
Now, as for `MeasureAndReset` I think the following could work (note the lack of “result” as a part of the statements):
```python
# in wire
class MeasureAndReset:
	measure_basis: ir.SSAValue = info.argument(OpType)
	reset_basis: ir.SSAValue = info.argument(OpType)
	out_wire: ir.ResultValue = info.result(WireType)

# in qubit
class MeasureAndReset:
	measure_basis: ir.SSAValue = info.argument(OpType)
	reset_basis: ir.SSAValue = info.argument(OpType)
	qubits: ir.SSAValue = info.argument(ilist.IListType[QubitType])
```

Currently, `RZ`, `RY`, and `RX` exist in the Stim dialect along with `MZ`, `MY`, `MX`, `MZZ`, `MYY`, and `MXX`. I think the Stim dialect we have would also need `MRZ`, `MRX`, and `MRY` (which I understand to be Measure in this pauli basis, then reset to whatever you consider |0> in that basis - although now I wonder would we have to let the user specify that? That seems overly complex. Like `MeasureAndReset(measure_basis = ..., reset_basis = ..., reset_value = ?)`) 

[...]

PR #265, opened 2025-05-13
MeasureAndReset/Reset changes from RFC

johnzl-777:
Implement changes agreed upon in #180


johnzl-777:
@weinbe58 I was able to get Reset to work with the PyQrack runtime but it does feel a bit odd now that it's a primitive op because I have to chuck it into Apply

src/bloqade/pyqrack/squin/runtime.py
```diff
@@ -41,6 +41,21 @@ def broadcast_apply(self, qubits: ilist.IList[PyQrackQubit, Any], **kwargs) -> N
...
+
+    @property
+    def n_sites(self) -> int:
+        return 1
+
+    def apply(self, *qubits: PyQrackQubit, adjoint: bool = False) -> None:
+        for qubit in qubits:
+            if not qubit.is_active():
+                continue
+            qubit.sim_reg.force_m(qubit.addr, 0)
```
weinbe58, line 56: Oh we have had issues with `force_m` so don't use it you can simply measure the qubit and if the measurement is `1` then just flip the state using an `x` operation on that qubit.

src/bloqade/squin/wire.py
```diff
@@ -95,21 +95,24 @@ def __init__(self, operator: ir.SSAValue, *args: ir.SSAValue):
...
     out_wire: ir.ResultValue = info.result(WireType)
 
 
 @statement(dialect=dialect)
-class Reset(ir.Statement):
+class MeasureAndReset(ir.Statement):
     traits = frozenset({lowering.FromPythonCall(), WireTerminator()})
     wire: ir.SSAValue = info.argument(WireType)
+    result: ir.ResultValue = info.result(types.Int)
+    out_wire: ir.ResultValue = info.result(WireType)
```
weinbe58, line 115: wait I thought in the issue this was supposed to be a terminator? that means it should take in.a qubit and only return the result of the measurement.

weinbe58 (changes requested):
Overall looks good! The PyQrack runtime for Reset is correct to overload `apply`, you just need to not use `force_m` which can throw an exception, see my comment.

Also I want to clarify what `MeasureAndReset` is not correct according to the conclusion of the rfc.

src/bloqade/squin/wire.py
johnzl-777: I'll be honest I'm a bit confused myself, what you're describing sounds like Measurement in the qubit dialect. All I really have from my notes is that MeasureAndReset was supposed to be removed from squin.qubit.

I also realize it's a bit confusing that the RFC conclusion was to remove MeasureAndReset from qubit dialect but right before that conclusion I say it should be removed from the wire dialect 😅

Roger-luo (approved):

johnzl-777:
Status update: reached agreement with @weinbe58 on what the actual change should be although @cduck might have something different in mind. Long story short my notes from the conclusion of the #180 Triage meeting seemed to be inaccurate. 

For the record, Phillip's confusion stems from the fact the statement above is labeled a terminator but does not obey the semantics of one (a mistake on my part). I know Phillip was unsure of the necessity of MR (MeasureAndReset) in wire owing to the fact Reset is now a primitive op in squin (abstracted away from either subdialect) but Casey seemed to have mentioned some benefit to having MR in wire.

cduck:
My understanding is the squin qubit dialect is for user programming and inter-operating with other dialects like QASM while the squin wire dialect is for circuit-level optimizations.  From this understanding, Measure and Reset need to be in the squin qubit dialect otherwise we cannot represent all circuits.

Also, for the squin wire dialect, some circuit optimizations like Rz gate elimination and unused qubit elimination need the context of the initial qubit state or measurement to perform the optimization.  I'll defer to others on the decision that these optimizations are better run on other dialects or left for a future date.

@weinbe58 @johnzl-777 @kaihsin

johnzl-777:
Just to put this on the record, the confusion around the conclusion of the triage meeting of #180 has been clarified by @cduck and @weinbe58 at today's sync meeting. The standard destructive measurement should definitely be a wire terminator and there should be no MeasureAndReset in EITHER squin wire or squin qubit dialects. 

CI seems happy with this PR, although one thing I would like to get more clarification on is how exactly the new NonDestructiveMeasurement statement is supposed to behave in the pyqrack runtime and in being rewritten to stim. Other than that, everything else is ready to go.

src/bloqade/squin/wire.py
```diff
@@ -95,23 +95,18 @@ def __init__(self, operator: ir.SSAValue, *args: ir.SSAValue):
...
     result: ir.ResultValue = info.result(types.Int)
 
 
 @statement(dialect=dialect)
-class MeasureAndReset(ir.Statement):
-    traits = frozenset({lowering.FromPythonCall(), WireTerminator()})
-    wire: ir.SSAValue = info.argument(WireType)
+class NonDestructiveMeasure(ir.Statement):
+    traits = frozenset({lowering.FromPythonCall()})
+    input_wire: ir.SSAValue = info.argument(WireType)
```
weinbe58, line 105: @cduck I think it would be good to understand the runtime behavior of this statement but @johnzl-777 please lets merge this PR and address it later. 

We can just error if you write this statement.

weinbe58 (approved):

PR #266, opened 2025-05-14
Add back returns kwarg to qasm2.loads / qasm2.loadfile

david-pl:


weinbe58 (approved):
this looks good to me. @Roger-luo what do you think?

Roger-luo (approved):
I'm not sure if we want to add back multiple returns but I think we can merge first and see if there's anyone complaining

david-pl:
> I'm not sure if we want to add back multiple returns

Was that supported before?

Roger-luo:
> Was that supported before?

yes, previous it accepts a tuple of `str`

PR #267, opened 2025-05-15
Fix address analysis for loop impl

david-pl:
Ran into some issues when trying to implement the scf methods for the fidelity analysis. There were some outdated methods, which I suppose were forgotten during an upgrade to a newer kirin version.

Also, the index can be an `IList` sometimes. Not sure if I fixed that case properly.

FYI, @weinbe58  @johnzl-777


src/bloqade/analysis/address/impls.py
```diff
@@ -73,8 +75,15 @@ def new_ilist(
 class PyIndexing(interp.MethodTable):
     @interp.impl(py.GetItem)
     def getitem(self, interp: AddressAnalysis, frame: interp.Frame, stmt: py.GetItem):
-        # Integer index into the thing being indexed
-        idx = interp.get_const_value(int, stmt.index)
+        stmt_index = stmt.index
+        if stmt_index.type.is_subseteq(ilist.IListType):
```
david-pl, line 79: Specifically, I'm not sure about this part here.

weinbe58: Wait when is an argument to `obj[...]` an ilist?

weinbe58:
Oh this might be why we are having an issue with the parallel rewrites #255

src/bloqade/analysis/address/impls.py
weinbe58: Wait but this has to be a bug then since `q[0]` the should lower to a getitem with the register `q`  being `obj` not `index`.

david-pl: Oh, and the code here is completely wrong. I'm now fetching the `BlockArgument` index, which is just the index in the block. That doesn't make sense.

How do you unwrap a `BlockArgument` to get the underlying value?

weinbe58: It should have a value in the analysis frame.

weinbe58: I think in this case the value for the block argument is`NoQubit`. The Analysis will fail for this example at the moment.

david-pl: @weinbe58 and that is fine? If so, that also means I can't make the fidelity analysis work for nested kernels such as this one, since we're tracking the atom survivability per qubit. Maybe we should just remove that for now? Then I wouldn't have to rely on the address analysis at all.

david-pl:
@weinbe58 I've reduced this PR to just fixing the outdated method calls in the analysis and adding some test cases for that. One test still fails (marked as xfail), but that is due to the bug in this issue: https://github.com/QuEraComputing/kirin/issues/408

I'd still go ahead and merge this as is, since it definitely fixes some issues.

Roger-luo (approved):

PR #273, opened 2025-05-16
Implement runtime for squin noise statements

david-pl:


src/bloqade/squin/noise/stmts.py
```diff
@@ -49,11 +49,18 @@ class PauliChannel(NoiseChannel):
...
 
 @statement(dialect=dialect)
 class QubitLoss(NoiseChannel):
     # NOTE: qubit loss error (not supported by Stim)
     p: ir.SSAValue = info.argument(types.Float)
+
+
+@statement(dialect=dialect)
+class StochasticUnitaryChannel(ir.Statement):
+    operators: ir.SSAValue = info.argument(ilist.IListType[OpType, NumOperators])
```
weinbe58, line 62: One issue here is what happens when the operators do not have the same number of sites. And also does this mean we might want to make `OpType` Generic with a type parameter for the operator size?

cc: @Roger-luo @johnzl-777 @kaihsin

david-pl: Well given the current noise statements, I don't think there can be a case where the number of sites is different. I could check this on runtime and throw an error if needed.

If we'd want to accommodate different sizes here, I think we can't have this share the same `OperatorRuntime`, since these all have a fixed number of sites. Also, what would be the use case?

> also what does this statement mean? randomly apply one of the unitary? What is the use case?

We're rewriting a bunch of noise statements to it so we can have a simple runtime. The noise statements are all represented by this single statement.

Roger-luo: > I could check this on runtime and throw an error if needed.

Yeah we should check it at runtime in this case, we can add an attribute to say it's validated to turn the checks off if we have validation later.

> We're rewriting a bunch of noise statements to it so we can have a simple runtime. The noise statements are all represented by this single statement.

I see! make sense!

david-pl:
@weinbe58 a couple of things to note:
* I've restricted the `Depolarize` channel to a single qubit only.
* The `QubitLoss` channel is not rewritten and has its own runtime, since it's something distinctly different from a stochastic unitary channel.
* I'm still not sure about the rewrite / implementation of the `PPError` channel as it's exactly the same as the `PauliError` channel: you provide an operator that is applied with a certain probability. Is that correct?

Ready for review!

src/bloqade/pyqrack/squin/noise/native.py
```diff
@@ -0,0 +1,72 @@
+import random
```
Roger-luo, line 1: why this is called native.py instead of runtime.py?

src/bloqade/pyqrack/squin/noise/native.py
```diff
@@ -0,0 +1,72 @@
...
+    @property
+    def n_sites(self) -> int:
+        n = self.operators[0].n_sites
+        for op in self.operators[1:]:
+            assert (
+                op.n_sites == n
+            ), "Encountered a stochastic unitary channel with operators of different size!"
+        return n
+
+    def apply(self, *qubits: PyQrackQubit, adjoint: bool = False) -> None:
```
Roger-luo, line 29: just double checking here pyqrack does not have a runtime like this and we need to implement the sampling ourselve?

david-pl: AFAIK pyqrack doesn't implement any noise sampling. Correct, @weinbe58 ?

FWIW I just checked the API of the `QrackSimulator` again and didn't find anything.

src/bloqade/squin/noise/stmts.py
```diff
@@ -26,34 +28,38 @@ class PPError(NoiseChannel):
...
-class PauliChannel(NoiseChannel):
-    # NOTE:
-    # 1-qubit 3 params px, py, pz
-    # 2-qubit 15 params pix, piy, piz, pxi, pxx, pxy, pxz, pyi, pyx ..., pzz
-    # TODO add validation for params (maybe during lowering via custom lower?)
-    n_qubits: int = info.attribute()
-    params: ir.SSAValue = info.argument(types.Tuple[types.Vararg(types.Float)])
-    result: ir.ResultValue = info.result(OpType)
+class SingleQubitPauliChannel(NoiseChannel):
+    params: ir.SSAValue = info.argument(types.Tuple[types.Float])
```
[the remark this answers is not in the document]
david-pl: yeah, it should be `IList` everywhere.

Roger-luo (approved):

Issue #275, opened 2025-05-16
Add `native` Noise dialect to QASM2 output as a comment.


weinbe58:
have a program like

```python
@qasm2.extended
def kernel():
    q = qasm2.Greg(3)
    native.pauli_channel([q[0]], px=0.01, py=0.02, pz=0.03)
    qasm2.h(q[1])

```

would lower to the following text:

```
OPENQASM 2.0;
#include "qelibc1.inc"

qreg[3] q;
// pauli_channel([q[0]], px=0.01, py=0.02, pz=0.03) 
h(q[1]);

```

cc: @stefan-o @lamq317 @Roger-luo

PR #277, opened 2025-05-16
Khwu/qasm2noise

kaihsin:
This PR address #275. Deps #279


kaihsin:
Note that this is depending on #279

src/bloqade/qasm2/dialects/noise.py
```diff
@@ -14,3 +16,33 @@ class Pauli1(ir.Statement):
...
+
+        qarg_str = (
+            f"{qarg.name.id}[{qarg.addr}]"
+            if isinstance(qarg, ast.Bit)
+            else f"{qarg.id}"
+        )
+
+        frame.body.append(
+            ast.Comment(
+                text=f"noist.Pauli1({px.value}, {py.value}, {pz.value}) {qarg_str}]"
```
david-pl, line 45:
```suggestion
                text=f"noise.Pauli1({px.value}, {py.value}, {pz.value}) {qarg_str}]"
```

src/bloqade/qasm2/emit/target.py
```diff
@@ -28,6 +29,7 @@ def __init__(
         allow_global: bool = False,
         custom_gate: bool = True,
         unroll_ifs: bool = True,
+        allow_noise: bool = False,
```
david-pl, line 32: Could we make that default to `True`? How bad would that be?

If you forget to set it, you'll get the error message
```python
ValueError: dialect py.ilist is not supported by DialectGroup([qasm2.expr, lowering.func, qasm2.core, func, scf, lowering.call, qasm2.uop, qasm2.indexing])
```
I think this will trip up most users and it's not clear at all what to do in order to fix it (in this case simply setting the `allow_noise=True`).

david-pl (approved):
Looks good, great usability improvement! Just two small comments.

src/bloqade/qasm2/emit/target.py
kaihsin: sg. lets make it default True

Roger-luo (approved):

Issue #227, opened 2025-05-02
Implement pyqrack methods for squin noise statements


david-pl:
LGTM can we have some basic tests on the constructors? Or maybe you could help @david-pl add the corresponding runtime in a new PR (which should test this too). I think would be nice to have it before next week's hackthon.

_Originally posted by @Roger-luo in https://github.com/QuEraComputing/bloqade-circuit/pull/211#pullrequestreview-2812423481_

weinbe58:
implement a single runtime statement for pyqrack and then rewrite common noise channels to this statement : 

```python
@statement(dialect=dialect)
class StochasticUnitaryChannel:
    operators: ir.SSAValue = argument(ilist.IListType[OpType, NumOperators]
    probabilities: ir.SSAValue = argument(ilist.IListType[float, NumOperators]
```

david-pl:
@weinbe58 working on #273 and now I'm wondering:
Do we even want the rewrite? In the squin runtime, we have to create a runtime object for operators anyway. So I could just as well give the statements a common interface, e.g. `get_operators` and `get_probabilities` getter methods (or just properties?) and use those to create a single common runtime object for all of them.

IMO, that would be easier, results in less code and is a bit clearer since the IR keeps the original statements.
Is there any reason why we should prefer the rewrite?

david-pl:
@weinbe58 to clarify what I mean, I drafted something up in #278. The runtime is the same, the method table is still just one method, basically: https://github.com/QuEraComputing/bloqade-circuit/blob/33f10736dbd6806ca334c088555f38e1490c3b99/src/bloqade/pyqrack/squin/noise.py#L14

Instead of the rewrite, I'd be adding common methods to the noise statements.

I'd argue that this approach is a bit simpler, but I'm sure you have some better insights, so I'd like to know which approach you prefer before continuing.

PR #278, opened 2025-05-17
Implement squin noise runtime without rewriting

david-pl:
Addresses #227, but uses an alternative approach without an additional rewrite.

Alternative to #273

### DIFF
--- a/src/bloqade/squin/groups.py
+++ b/src/bloqade/squin/groups.py
@@ -3,7 +3,7 @@ from kirin.prelude import structural_no_opt
 from kirin.dialects import ilist
 from kirin.rewrite.walk import Walk
 
-from . import op, wire, qubit
+from . import op, wire, noise, qubit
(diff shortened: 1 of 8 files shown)


src/bloqade/squin/groups.py
```diff
@@ -44,3 +44,11 @@ def run_pass(method):
         py_mult_to_mult_pass(method)
 
     return run_pass
+
+
+@ir.dialect_group(kernel.add(noise.dialect))
```
Roger-luo, line 49: maybe just put noise dialect inside kernel? like how to define `qasm2.extended`?

david-pl: Yes, this was just because I had a separate decorator in #273 with an additional rewrite. I'll add it to the kernel directly. But first, I'll need to see whether we're going with #273 or this version.

david-pl:
After discussion with @weinbe58 we decided to go with #273 instead.

Issue #276, opened 2025-05-16
`import *` causing import errors


cduck:
Example caused by this line: https://github.com/QuEraComputing/bloqade-circuit/blob/23c995b40b91f155baee40b4a7312eeb5e37baad/src/bloqade/stim/__init__.py#L5
```python
import bloqade.stim.emit as module1
import bloqade.stim.dialects.collapse.emit as module2

print(module1.__name__)  # Prints: bloqade.stim.dialects.collapse.emit
print(module2.__name__)  # Prints: bloqade.stim.dialects.collapse.emit
```

In this specific case, `bloqade/stim/dialects/collapse/emit.py` imports enough names from `bloqade/stim/emit.py` to avoid errors on typical imports.  But I got this weird error:

```python
>>> import bloqade.stim.emit.stim as something
---------------------------------------------------------------------------
ImportError                               Traceback (most recent call last)
Cell In[9], line 1
----> 1 import bloqade.stim.emit.stim as something

ImportError: cannot import name 'stim' from 'bloqade.stim.dialects.collapse.emit' (/Users/cduckering/docs/dev/qec-arch/qec_exec/.venv/lib/python3.10/site-packages/bloqade/stim/dialects/collapse/emit.py)
```

-----

Solutions:
- All modules that are intended to be used with `import *` should include the `__all__ = ['var1', 'var2', ...]` magic list to restrict what `import *` includes.
- Or don't use `import *`.

kaihsin:
It should. I'll double check

PR #279, opened 2025-05-19
Clean up import path

kaihsin:
This PR address #276, remove all `from xxx import *`

General pattern:

1. to access the Statement definition: 

```python 
from stim.dialects import  collapse, auxiliary, gate

gate.CX # the statement def 
auxiliary.ObservableInclude # the statement def
```

2. for general user access, they should not touch statement, so:
```python
from bloqade import stim

stim.cx # wrapper 
stim.observable_include # wrapper
```


Roger-luo:
So the issue was due to * imports other names not just the statements?

Roger-luo (approved):

PR #280, opened 2025-05-20
enable squin -> stim tests and fix imports from latest stim module shuffle

johnzl-777:
@ChenZhao44 brought to my attention a couple things with my squin to stim tests and the existing infrastructure while he was developing things. 

The latest import shuffle from #279 does break my existing work but is very easy to fix. On top of that it turns out the test file I made was never actually run so the import change would have flown under the radar if it had not been for Chen's poking around (: 

I went ahead and commented out the excess printing just so running coverage locally/in CI doesn't end up spitting a bunch of stuff out.


johnzl-777:
Changes made here will need to be present to complete #265 , this PR is no longer needed

PR #288, opened 2025-05-22
Add parallelization function using cirq and linear programming

jon-wurtz:


jon-wurtz:
Create a new helper function `parallelize`, which uses DAG representations of static circuits to find parallelism. The strategy uses linear programming, where each gate in the circuit is assigned a "time", where DAG constraints take the form `t2-t1>1` in a linear constraint solver. The program is solved all at once with the `qpsolvers` package, for which I've added a few convenient wrappers to not have to deal with basis analysis.

Issue #283, opened 2025-05-20
interop with cirq


Roger-luo:
I think given most of us are using cirq here and there, we should start with taking cirq circuit as an input of the compiler directly. This would allow users from cirq passing their circuit program to bloqade. I don't have a concrete proposal on what the interface gonna look like, here are a few possible options:

- similar to QASM2, we provide a `bloqade.cirq.load_circuit(Circuit)` function to convert a `cirq.Circuit` into `Method`
- proposed by @weinbe58 treat cirq circuit as an operator.

PR #294, opened 2025-05-27
Implement lowering a cirq.Circuit to squin

david-pl:
Addresses #283 

For now, I went with the syntax `squin.load_circuit(circuit: cirq.Circuit)`.

So you can e.g. do

```python
q0 = cirq.NamedQubit("q0")
q1 = cirq.NamedQubit("q1")

circuit = cirq.Circuit(
    cirq.H(q1),
    cirq.X(q0).controlled_by(q1),
    cirq.Rx(rads=math.pi / 4).on(q0).controlled_by(q1),
    cirq.measure(q0, q1),
)

main = squin.load_circuit(circuit)
```

~~@Roger-luo @weinbe58 one open question here: cirq support arbitrary powers of gates, e.g. `cirq.H(q0)**0.123`. Should we add a `Pow` statement to squin to properly support this? I think pyqrack can also handle it, but I'll have to check again. Also, not sure what this should look like when running on hardware, but we can still error in the runtime.~~

**Edit**: turns out arbitrary powers are quite easy since we can just use the cirq helper `in_su2` to get the corresponding rotation and lower that. Also, for multi-qubit gates they usually mean "apply a controlled single qubit gate with an exponent", which again corresponds to a rotation. Thanks @kaihsin for pointing to the helper method.

## TODOs:

- [x] support (arbitrary) powers of operators?
- ~~[ ] Lowering of noisy circuits~~ done, except #301 
- [x] Add missing gates (e.g. three-qubit controls are still missing)
- [ ] Make sure the set of supported gates is complete
- ~~ [ ] Optionally return the quantum register (and possibly other things) ~~ see #302 
- ~~ [ ] Optionally pass in the quantum register as argument to the method ~~ see #302
- ~~[ ] Support classical control by lowering to `if` ~~ see #303

### DIFF
--- a/src/bloqade/squin/cirq/lowering.py
+++ b/src/bloqade/squin/cirq/lowering.py
@@ -282,7 +282,9 @@ class Squin(lowering.LoweringABC[CirqNode]):
 
         # TODO: do we need a dedicated noise stmt for this? Using PauliError
         # with this basis feels like a hack
-        noise_channel = noise.stmts.PauliError(basis=r.result, p=p.result)
+        noise_channel = state.current_frame.push(
+            noise.stmts.PauliError(basis=r.result, p=p.result)
+        )
 
         return noise_channel
 
--- a/test/squin/cirq/test_cirq_to_squin.py
+++ b/test/squin/cirq/test_cirq_to_squin.py
@@ -5,6 +5,9 @@ import pytest
 
 from bloqade import squin
 from bloqade.pyqrack import DynamicMemorySimulator
+from bloqade.squin.noise.rewrite import RewriteNoiseStmts
+
+rewrite_noise_pass = RewriteNoiseStmts(squin.kernel)
 
 
 def basic_circuit():
@@ -149,6 +152,8 @@ def test_circuit(circuit_f, run_sim: bool = False):
     kernel = squin.load_circuit(circuit)
 
     kernel.print()
+
+    rewrite_noise_pass(kernel)
 
     # make sure we produce a valid kernel by running it
     sim = DynamicMemorySimulator()


jon-wurtz:
When transpling to squin, could you also give an option that moments are transpiled correctly to preserve ordering? This follows the PR that I made a few days ago https://github.com/QuEraComputing/bloqade-circuit/pull/288 where I parallelize, represented by cirq moments.

src/bloqade/squin/cirq/lowering.py
```diff
@@ -0,0 +1,275 @@
...
+        self, state: lowering.State[CirqNode], node: cirq.Circuit
+    ) -> lowering.Result:
+        for moment in node:
+            state.lower(moment)
+
+    def visit_Moment(
+        self, state: lowering.State[CirqNode], node: cirq.Moment
+    ) -> lowering.Result:
+        for op_ in node.operations:
+            state.lower(op_)
```
Roger-luo, line 127: maybe in a separate PR, to address @jon-wurtz 's comment, we should: 

- lower moment with same gate into `broadcast(op, ...)`
- lower moment with different gate into `parallel` statement (which will be added on Kirin side)

to facilitate this we may want to just add a `Moment` statement as an IR to temperarily perserve this highlevel structure coming from cirq and then implement a rewrite from `Moment` to squin.

Roger-luo (approved):
I think the PR itself LGTM. I left a comment to improve in the future PR. We should get this basic API in first.

src/bloqade/squin/cirq/lowering.py
```diff
@@ -0,0 +1,287 @@
...
+        with state.frame(
+            [stmt],
+            globals=globals,
+            finalize_next=False,
+        ) as frame:
+            # NOTE: create a global register of qubits first
+            # TODO: can there be a circuit without qubits?
+            n_qubits = cirq.num_qubits(self.circuit)
+            n = frame.push(py.Constant(n_qubits))
+            self.qreg = frame.push(qubit.New(n_qubits=n.result))
```
jon-wurtz, line 85: Might it be better to have the register be an input to the kernel instead of inside of the kernel? Otherwise, the kernel generated cannot be used as a subroutine elsewhere. Likewise, returning the qubit register. See https://github.com/QuEraComputing/bloqade/issues/249.

david-pl: @jon-wurtz see my comment in the RFC issue (let's leave the discussion in on place). Ideally we support both, loading a "standalone" circuit as well as passing the register as an argument, controlled by a kwarg.

david-pl:
@weinbe58 @Roger-luo I think we need to add a new operator statement in order to support the `GeneralizedAmplitudeDampingChannel` from cirq.

Basically, this channel maps |0> to |1> and vise-versa, but at different probabilities. So essentially I need both processes separately. Mapping from |1> to |0> can be done via `Reset`, but there is no statement that maps |0> to |1> (the projector can potentially lead to 0 amplitude of the state).

Roger-luo:
> @jon-wurtz classical control is still missing. That needs to be lowered to an if statement. I updated the TODO list in the PR description.

can we limit this PR to have the basic only? otherwise it will take forever to merge...

>  I think we need to add a new operator statement in order to support the GeneralizedAmplitudeDampingChannel from cirq.

I think that's ok? I don't see a reason why no.

src/bloqade/squin/cirq/lowering.py
Roger-luo: or a Moment should be rewrite into something like a `Parallel` with a region?

PR #305, opened 2025-05-28
Adding native operators dialect. 

### LINKED ISSUE #297: Split out Native gate set into separate dialect.
Talking with @kaihsin we discussed using another dialect to constrain the operators allowed downstream in `bloqade-shuttle `and lower levels. E.g. we should not allow composite operators inside of `bloqade-shuttle`

weinbe58:
closes #297

### DIFF
--- a/src/bloqade/squin/__init__.py
+++ b/src/bloqade/squin/__init__.py
@@ -4,6 +4,7 @@ from . import (
     noise as noise,
     qubit as qubit,
     lowering as lowering,
+    _typeinfer as _typeinfer,
 )
 from .groups import wired as wired, kernel as kernel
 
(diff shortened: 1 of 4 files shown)


src/bloqade/squin/native/_wrapper.py
```diff
@@ -0,0 +1,46 @@
...
+
+    Returns:
+        Op: The resulting operator.
+    """
+    ...
+
+
+@lowering.wraps(R)
+def r(
+    axis_angle: float,
```
kaihsin, line 20: missing rotation_angle

kaihsin (changes requested):

PR #306, opened 2025-05-31
move control gate mapping outside function for external usage

johnzl-777:
I've found that it would be nice to have the dictionary for mapping between squin 1Q operators and supported Stim dialect control gate operators to be accessible outside the function, especially for other squin -> stim related rewrite work.

Not a blocker for anything, just a minor improvement in convenience.


johnzl-777:
Moved into #310

Issue #270, opened 2025-05-15
`parallel.u` to `glob.u` rewrite/pass.


weinbe58:
Currently there is no way to do this we should definitely have this for better noise modeling. 

The idea here is to simply run the qubit analysis and then look for Parallel statements that use an entire reigster and replace that with a global gate statement.

PR #309, opened 2025-06-02
Adding reg as a value inside of AddressQubit element. 

weinbe58:
I want to add this because on the case of trying to implement the rewtie in #270 you need to check if all qubits come from the same register value if they are inside an `AddressTuple` lattice element. Here the `AddressQubit` as a reference to the register object it comes from and so its possible to figure out just from `AddressTuple` locally if all the elements of an `AddressReg` are inside the `AddressTuple`.


weinbe58:
Actually we don't need this if you just look at the SSA you can obtain this information

PR #310, opened 2025-06-03
Improvements to squin-related infrastructure

johnzl-777:
These were changes I originally planned to make in separate PRs but considering how useful they are for other work along with the fact it would get a bit messy having so many very small PRs I thought it best to consolidate them here.

Changes include:
- support for site analysis to work with kirin scf dialect
- enabling the squin to stim control gate mapping to be externally accessible


src/bloqade/squin/rewrite/remove_dangling_qubits.py
weinbe58, line 1: Why is this necessary? `DeadCodeElimination` should also remove these statements and if it doesn't it is a bug.

src/bloqade/squin/rewrite/wrap_analysis.py
```diff
@@ -70,3 +60,34 @@ def rewrite_Statement(self, node: ir.Statement) -> RewriteResult:
...
+
+    def wrap(self, value: ir.SSAValue) -> bool:
+        address_analysis_result = self.address_analysis[value]
+
+        if value.hints.get("address"):
+            return False
+        else:
+            value.hints["address"] = AddressAttribute(address_analysis_result)
+
+        return True
```
weinbe58, line 77:
```suggestion
    def wrap(self, value: ir.SSAValue) -> bool:
        if value.hints.get("address") is not None:
            return False
        
        address_analysis_result = self.address_analysis[value]
        value.hints["address"] = AddressAttribute(address_analysis_result)

        return True
```

nitpick on style.

src/bloqade/squin/rewrite/wrap_analysis.py
```diff
@@ -40,22 +41,11 @@ def print_impl(self, printer: Printer) -> None:
...
-            return False
-        else:
-            value.hints["address"] = AddressAttribute(address_analysis_result)
-            value.hints["sites"] = SitesAttribute(op_site_analysis_result)
-
-        return True
+        pass
 
     def rewrite_Block(self, node: ir.Block) -> RewriteResult:
         has_done_something = False
```
weinbe58, line 51: do you need this? You should just use a `Walk` and only implement the `rewrite_Statement` method since the analysis results are referenced by the SSAValues.

src/bloqade/squin/rewrite/wrap_analysis.py
```diff
@@ -70,3 +60,34 @@ def rewrite_Statement(self, node: ir.Statement) -> RewriteResult:
             if self.wrap(result):
                 has_done_something = True
         return RewriteResult(has_done_something=has_done_something)
```
weinbe58, line 62:
```suggestion
        return RewriteResult(has_done_something=any(self.wrap(result) for result in node.results)
```
could be a one-liner

src/bloqade/squin/rewrite/wrap_analysis.py
```diff
@@ -70,3 +60,34 @@ def rewrite_Statement(self, node: ir.Statement) -> RewriteResult:
...
+
+    def wrap(self, value: ir.SSAValue) -> bool:
+        op_site_analysis_result = self.op_site_analysis[value]
+
+        if value.hints.get("sites"):
+            return False
+        else:
+            value.hints["sites"] = SitesAttribute(op_site_analysis_result)
+
+        return True
```
weinbe58, line 93: same here

weinbe58 (changes requested):
Overall looks OK, I have two concerns:

1. Why is there a `RemoveDanglingQubits` rule? this should be captured by DeadCodeElimination or am I missing something.
2. The `WrapSquinAnalysis` base claas doesn't need to implement both `rewrite_Block` and `rewrite_Statement` if you use it with a `Walk` because `Walk` will apply both rules so you should either implement one or the other not both.

weinbe58 (approved):
OK talked with @johnzl-777 we're all set

PR #315, opened 2025-06-11
Support cirq classical control when lowering a circuit to squin

david-pl:
Still need to support classical control with measurements on multiple qubits resulting in an ilist.


src/bloqade/squin/cirq/lowering.py
```diff
@@ -150,10 +150,79 @@ def lower_measurement(
...
+                    sym_name="bool_op_or",
+                    arg_names=[],
+                    dialects=self.dialects,
+                    code=f_code,
+                )
+                f_const = state.current_frame.push(py.constant.Constant(fn))
+                init_val = state.current_frame.push(py.Constant(False)).result
+                condition = state.current_frame.push(
+                    ilist.Foldl(f_const.result, measurement_outcome, init=init_val)
+                ).result
```
david-pl, line 197: @Roger-luo Please take a look at this part here specifically. I'm not sure if I'm adding the function here correctly.

Roger-luo (approved):

PR #326, opened 2025-06-19
RFC: More precise depolarize statements in squin

johnzl-777:
This occurred to me when this conversation between @cduck and @kaihsin (https://github.com/QuEraComputing/bloqade-circuit/pull/322#pullrequestreview-2937116152) on #322 prompted me to start working on a rewrite from squin's `Depolarize` to the stim dialect equivalents.

The problem is that squin's Depolarize is a bit ambiguous as to how to map onto the stim depolarize statements. 

For context, squin's Depolarize looks like this:

```python
@statement(dialect=dialect)
class Depolarize(NoiseChannel):
    """
    Apply depolarize error to qubit
    """

    p: ir.SSAValue = info.argument(types.Float)
```

Whereas Stim has `DEPOLARIZE1` (one qubit depolarization) and `DEPOLARIZE2` (two qubit depolarization). 

To show the ambiguity I run into, let's say I see the following:

```
squin.broadcast(depolarize(0.1), qubits = [1, 2])
```
how do I know if this is supposed to map to DEPOLARIZE1 or DEPOLARIZE2? 

Or

```
squin.broadcast(depolarize(0.01), qubits = [1,2,3,4])
```
?

To an extent, I could infer something like if the targets are odd in number it _must_ be DEPOLARIZE1 but if I have an even number it could be either/or. 

I propose letting squin have a 1:1 mapping with the two depolarize statements that stim has for this purpose. 

As for #322 I think it should either be merged in for now considering @ChenZhao44 and @liupengy19 will need the existing qubit loss functionality soon OR this proposal (or sufficient counter proposal) gets accepted and I can resolve the conversation mentioned in the beginning of this RFC.


weinbe58:
The convention is that for `broadcast(op: Op, qubits: ilist[Qubit, m])` the operator is applied to a list of qubits with a length equal to `m=n*nsites` where `n=1,2,3,4,...` and `nsites` is the number of sies of `op` the operator is applied to `qubits[i*nsites:(i+1)*nsites]` for `i=0, ..., n-1`.

johnzl-777:
Whoops, looks like I didn't account for the noise statement rewrite but that should be an easy fix

weinbe58:
Why don't we just add a `nsites` attribute to depolarize?

johnzl-777:
@ehua7365 has successfully added the necessary infrastructure to resolve my original concern for the RFC in #368 (:

Issue #324, opened 2025-06-18
Bug in NSitesAnalysis with the stdlib functions.


weinbe58:
One of the stdlib functions is not returning the correct value for the NSitesAnalysis. MWE:

```python

from bloqade import squin
from bloqade.squin.analysis.nsites import NSitesAnalysis

@squin.kernel
def main():
    qreg = squin.qubit.new(4)
    
    squin.qubit.apply(squin.op.h(), qreg[0])
    squin.qubit.apply(squin.op.cx(), qreg[0], qreg[1])
    squin.qubit.apply(squin.op.cx(), qreg[0], qreg[2])
    squin.qubit.apply(squin.op.cx(), qreg[0], qreg[3])

analysis_frame, ret= NSitesAnalysis(main.dialects).run_analysis(main,  no_raise=False)
main.print(analysis=analysis_frame.entries)

```

prints out:

```
func.func main() -> !py.NoneType {
  ^0(%main_self):
  │   %main_self --> NoSites()
  │    %0 = py.constant.constant 4 : !py.int # ---> NoSites()
  │ %qreg = squin.qubit.new(n_qubits=%0) : !py.IList[!py.Qubit, Literal(4,int)] # ---> NoSites()
  │    %1 = squin.op.h() : !py.Op # ---> NumberSites(sites=1)
  │    %2 = py.constant.constant 0 : !py.int # ---> NoSites()
  │    %3 = py.indexing.getitem(%qreg, %2) : !py.Qubit # ---> NoSites()
  │    %4 = py.ilist.new(values=(%3)){elem_type=!Any} : !py.IList[!py.Qubit, Literal(1,int)] # ---> NoSites()
  │         squin.qubit.apply(operator=%1, qubits=%4)
<<<<<<< missing analysis
  │    %5 = func.invoke cx() : !py.Op maybe_pure=True # ---> NoSites()
<<<<<<<
  │    %6 = py.constant.constant 0 : !py.int # ---> NoSites()
  │    %7 = py.indexing.getitem(%qreg, %6) : !py.Qubit # ---> NoSites()
  │    %8 = py.constant.constant 1 : !py.int # ---> NoSites()
  │    %9 = py.indexing.getitem(%qreg, %8) : !py.Qubit # ---> NoSites()
  │   %10 = py.ilist.new(values=(%7, %9)){elem_type=!Any} : !py.IList[!py.Qubit, Literal(2,int)] # ---> NoSites()
[...]

PR #327, opened 2025-06-20
Allow nsite analysis to work with invoke

johnzl-777:
Added a test that pretty much mirrors the MWE @weinbe58 showed in #324 , incorporating the fix that @david-pl was working on (: 

I also took the liberty to simplify a lot of old tests a bit using the kernel decorator (which either didn't exist when I made the tests OR I wasn't aware of it until after the tests were made 😅 )


weinbe58 (approved):
LGTM

PR #340, opened 2025-06-27
adding Wired statement for regionalized wire semantics

weinbe58:
The idea here is to replace phase out `wrap` and `unwrap` with regionalized semantics. In this case the inputs to the `wired` statement are the qubits to be converted into wires as block arguments that have been unwrapped and when executing the region the last wires are automatically wrapped back to their respective qubits.


src/bloqade/squin/wire.py
```diff
@@ -50,6 +50,14 @@ class Unwrap(ir.Statement):
     result: ir.ResultValue = info.result(WireType)
 
 
+@statement(dialect=dialect)
+class Wired(ir.Statement):
+    traits = frozenset()
+
+    qubits: tuple[ir.SSAValue, ...] = info.argument(QubitType)
+    body: ir.Region = info.region(multi=False)
```
Roger-luo, line 58: I think this needs to enable `multi` otherwise you lost the ability to have a complete CFG - which is completely legit.

Roger-luo (changes requested):
I think this PR is missing a few more specific things.

Roger-luo:
there should be a canonicalize rewrite associated with this too - when the region is empty or the argument list is empty, the statement should be inlined.

weinbe58:
@Roger-luo We probably should not allow for return statements inside this region right?

weinbe58:
Also @Roger-luo do we support return values for this statement, e.g. measurement results or something of the like?

PR #346, opened 2025-07-01
Use integers for tagging to avoid rounding eerrors

weinbe58:
@david-pl This shuold fix your CI locally I was also hving similar problems.


david-pl:
@weinbe58 yep, that did the trick. Tests work for me locally again. Thanks!

PR #358, opened 2025-07-03
add u3 to clifford rewrite rule to squin to stim pass

johnzl-777:
cc: @ehua7365


david-pl:
@kaihsin Does this need to be backported to v0.4?

PR #369, opened 2025-07-08
Add heuristic noise model in cirq

david-pl:
Credits to @tcochran-quera and @lamq317 for the implementation.

For now, this is mostly just a simple port of the internally developed heuristic noise model in cirq.
There are some potential improvements I see:

- The `get_two_zoned_noisy_circ` could use a custom class that inherits from `cirq.NoiseModel` to organize the code and make it a bit cleaner.
- The `utils.py` could also use some clean up. There might still be some functions that are actually not needed that I missed. Also we may want to think about moving the functions to classes to organize the code a bit more.

I didn't do any of that so far for a few reasons

- Mainly, this would break our internally developed notebooks, without adding any benefits besides a bit of a cleaner codebase (no new features or performance gains).
- We can still do it in the future and it's often useful to make changes one at a time.

@tcochran-quera @lamq317 @jon-wurtz please have a look at this PR and try it out. Let me know if you find things that can be improved / cleaned up / unified with existing functionality / removed.

One side note: this PR also adds the lowering to `squin`, meaning you can use `squin.cirq.load_circuit` in order to obtain a squin kernel from a (noisy) circuit.

**Edit:** For posterity, adding the summary of the RFC posted in a comment below here as well

# RFC summary

I'd like to summarize what we've discussed so far. Hopefully, this lets us conclude the "RFC" part of this PR and move on to finalizing things.

- **General API**: There will be a single, high-level transformation function `bloqade.cirq_utils.noise.transform_circuit(circuit, model)`, which essentially does two things:
  1. Converts the circuit to one using the native gate set only using `cirq.optimize_for_target_gateset`.
  2. Then applies the noise from the model (using what is essentially `circuit.with_noise(model)`). Idle qubits that are part of the non-native circuit will still be subjected to the move noise (this was the discussion on init noise we had).
- **Two-zone noise**:
  - The two-zone noise transformation is refactored into a `cirq.NoiseModel`. This allows both us and the user to use `circuit.with_noise` with the respective model. It should be noted, however, that the noise models are only compatible with native circuits. Also, there might be some edge cases due to initialization noise. This needs to be clearly documented in the respective doc strings.
  - The utility functions for the two-zone noise model are mostly left as-is. However, I'd consider them as internal helper functions only, so we should be able to make changes later on without breaking things for users.
  - The previously used function `get_two_zoned_noisy_circ` will be removed. Users should use `transform_circuit` instead.
- **One-zone noise**:
  - The interleaving is only part of one-zone noise models and as such will be moved into those noise models. Again, this makes it possible for the user to also just use `circuit.with_noise` so long as they keep the restriction to native gates in mind.
  - The `transform_to_noisy_one_zone_circuit` will also be removed and replaced by `transform_circuit`.


src/bloqade/cirq_utils/noise/utils.py
```diff
@@ -0,0 +1,815 @@
...
+
+            result["angles"].append(angles)
+        elif isinstance(op.gate, cirq.CZPowGate):  # CZ gate in Cirq
+
+            result["cz"].append(op.qubits)
+
+    return result
+
+
+def get_unique_qubits(list1, list2):
```
lamq317, line 709: This function can be removed, not used in any of the models

src/bloqade/cirq_utils/noise/utils.py
```diff
@@ -0,0 +1,815 @@
...
+        model = GeminiOneZoneNoiseModel()
+
+    system_qubits = circuit.all_qubits()
+    # Transform to CZ + PhasedXZ gateset.
+    if to_target_gateset:
+        with warnings.catch_warnings():
+            warnings.simplefilter("ignore")
+            # TODO: parallelize does this optimization too;
+            # check if we can skip this one here if parallelize is True
+            native_circuit = cirq.optimize_for_target_gateset(
```
tcochran-quera, line 78: There is a bit of magic in the cirq optimizers, but maybe for consistency I should try to composite function below here: `optimize_circuit_to_cz_gate_set`

src/bloqade/cirq_utils/noise/utils.py
```diff
@@ -0,0 +1,815 @@
...
+    Args:
+        moment: A cirq.Moment object.
+
+    Returns:
+        A list of tuples, where each tuple contains the qubits acted on by a gate in the moment.
+    """
+
+    list_qubs = [op.qubits for op in moment.operations]
+
+    # return [item for tup in list_qubs for item in tup]
```
tcochran-quera, line 167: dead code

test/cirq_utils/noise/test_noisy_ghz.py
```diff
@@ -0,0 +1,202 @@
...
+
+        # Take the trace and square it
+        fidelity_value = np.trace(sqrt_product)
+        return np.real(fidelity_value) ** 2
+
+    trans_circuits = transform_to_noisy_one_zone_circuit(create_On_ghz_circuit(4))
+    compressed_circuit = trans_circuits[0]
+    noisy_circuit = trans_circuits[1]
+
+    create_On_ghz_circuit(4)
```
tcochran-quera, line 66: What is this line doing?

tcochran-quera, line 66: What are these lines doing?

test/squin/cirq/test_cirq_to_squin.py
```diff
@@ -127,6 +129,17 @@ def noise_channels():
     )
 
 
+def depolarizing_channels():
+    q = cirq.LineQubit.range(2)
+
+    return cirq.Circuit(
+        cirq.depolarize(0.1)(q[0]),
+        cirq.asymmetric_depolarize(p_x=0.1)(q[0]),
+        cirq.asymmetric_depolarize(error_probabilities={"XY": 0.1})(*q),
```
tcochran-quera, line 138: Great find! I think we can replace replace the TwoQubitPauli class with this.

src/bloqade/cirq_utils/noise/custom_gates.py
```diff
@@ -0,0 +1,63 @@
+import cirq
+import numpy as np
+
+
+class TwoQubitPauli(cirq.Gate):
```
tcochran-quera, line 5: I think we can replace this with something like: `cirq.asymmetric_depolarize(error_probabilities={"XY": 0.1})(*q)`

tcochran-quera:
Maybe it would be helpful to split some of higher level functions that a user might directly interact with in `cirq_util/noise/utils.py` into a seperate file. We may also consider writing a common wrapper to apply noise and have `zone` as an argument. Along these lines, I think it would be nice if we could adopt a 'conflict graph' approach for the 2-zone model and a 'swaps' approach for the 1-zone model.

tcochran-quera (commented):
Closing my review for now.

plquera:
can I add a comment here about uniformizing syntax and workflow between the one-zone and two-zone models? Users should feed circuits and change model parameters in the same way for both, even if the models differ in how they build the circuits in the end.

src/bloqade/cirq_utils/noise/utils.py
david-pl: @tcochran-quera I'll leave that to you then, okay?

david-pl:
> Maybe it would be helpful to split some of higher level functions that a user might directly interact with in `cirq_util/noise/utils.py` into a seperate file. We may also consider writing a common wrapper to apply noise and have `zone` as an argument. Along these lines, I think it would be nice if we could adopt a 'conflict graph' approach for the 2-zone model and a 'swaps' approach for the 1-zone model.

I agree, the utils file should be reorganized. That common wrapper depends a bit on whether we want to refactor the two-zone noise into a proper noise model. I think we can have noisy moments that depend on previous ones by directly overloading the `noisy_moments` method rather than the single moment `noisy_moment` one, but I'll have to check in detail.

The last two points (conflict graph for two zone, swaps for one zone) sound like additions that could be done in the future. I'd suggest we spend some time here improving what we have so we get a good basis here and leave those for future PRs. What do you think?

tcochran-quera:
wrt [008f94a](https://github.com/QuEraComputing/bloqade-circuit/pull/369/commits/008f94a6b9135c5f930d2fe4d2820d25f4196195)

@lamq317 pointed out that the interleaving procedure is not compatible with the two zone noise model (it will introduce extra noise moments). To address this, I created a new noise class `GeminiOneZoneNoiseModel_v1`, which performs the interleaving inside `.noisy_moments` and I also added a `transform_circuit_v1`. I also added a test file to verify that `transform_circuit` and `transform_circuit_v1` produce the same result.

If we agree on this approach, we should similarly implement the interleaving for each of the OneZone noise models... suggesting that the interleaving code should probably be dried. I'll wait for comments before diving into that. @david-pl 

ps. I also updated the name of the `TwoZoneNoiseModel` to `GeminiTwoZoneNoiseModel` for consistency on @lamq317 suggestion.

david-pl:
@tcochran-quera great, thanks! This looks good, IMO. I have some questions:

- **Initialization noise**: just to follow up on our discussion regarding this, passing the `system_qubits` into the `noisy_moments` now alleviates the need for adding a "fake" identity gate into the circuit and then removing it. Is that correct?
- **Interleaving**: thanks for catching that @lamq317, I was a bit too quick to re-use the transformation function it seems. I think your approach here is good, @tcochran-quera. If something is not compatible across all the noise models when adding noise to the circuit, it should simply be part of the noise model rather than the higher-level function. At least, that's my opinion. So, basically, I fully agree with your approach here. You could think about implementing it only once in a base class e.g. `GemineOneZoneBaseClass(cirq.NoiseModel)` and then inheriting from that for all the one-zone models. This works so long as the interleaving is compatible across all one-zone noise models.

david-pl:
# RFC summary

I'd like to summarize what we've discussed so far. Hopefully, this lets us conclude the "RFC" part of this PR and move on to finalizing things.

- **General API**: There will be a single, high-level transformation function `bloqade.cirq_utils.noise.transform_circuit(circuit, model)`, which essentially does two things:
  1. Converts the circuit to one using the native gate set only using `cirq.optimize_for_target_gateset`.
  2. Then applies the noise from the model (using what is essentially `circuit.with_noise(model)`). Idle qubits that are part of the non-native circuit will still be subjected to the move noise (this was the discussion on init noise we had).
- **Two-zone noise**:
  - The two-zone noise transformation is refactored into a `cirq.NoiseModel`. This allows both us and the user to use `circuit.with_noise` with the respective model. It should be noted, however, that the noise models are only compatible with native circuits. Also, there might be some edge cases due to initialization noise. This needs to be clearly documented in the respective doc strings.
  - The utility functions for the two-zone noise model are mostly left as-is. However, I'd consider them as internal helper functions only, so we should be able to make changes later on without breaking things for users.
  - The previously used function `get_two_zoned_noisy_circ` will be removed. Users should use `transform_circuit` instead.
- **One-zone noise**:
  - The interleaving is only part of one-zone noise models and as such will be moved into those noise models. Again, this makes it possible for the user to also just use `circuit.with_noise` so long as they keep the restriction to native gates in mind.
  - The `transform_to_noisy_one_zone_circuit` will also be removed and replaced by `transform_circuit`.
- The conflict graph coloring algorithm could be replaced by the networkx one as @jon-wurtz pointed out. This can easily be done in the future too, however, so it's optional for now.


@tcochran-quera @lamq317 @jon-wurtz @plquera are we all agreed on this? FYI, I'll count not responding as silent agreement 😉

tcochran-quera:
@lamq317 @david-pl I noticed that `parallelize.transpile` function adds identity `cirq.PhasedXZ` gates. The one-zone model then treats this as a local gate.

lamq317:
@david-pl , `transform_to_qasm_u_gates` is originally there, If I remember correctly, for interoperability with the qasm2 dialect in bloqade, back when I was running simulations with pyqrack.  I think it is no longer necessary if we can directly lower the CZ+PhXZ gate set to squin.

src/bloqade/cirq_utils/noise/conflict_graph.py
```diff
@@ -0,0 +1,144 @@
...
+                    for connected_node in connected_nodes:
+                        if connected_node in move_schedule[color]:
+                            has_colored_neighbor = True
+                            break
+                        else:
+                            continue
+                    mover_limit_reached = len(move_schedule[color]) >= mover_limit
+                    if (has_colored_neighbor is False) and (
+                        mover_limit_reached is False
+                    ):
```
tcochran-quera, line 128: ): # if not, we color the node with this color.

src/bloqade/cirq_utils/noise/conflict_graph.py
```diff
@@ -0,0 +1,144 @@
...
+                        move_schedule[color].add(node)
+                        colored = True
+                        colored_nodes.add(node)
+                        colored_nodes.add(partner_node)
+                        break
+                if colored is False:
+                    move_schedule[len(move_schedule)] = {node}
+                    colored = True
+                    colored_nodes.add(node)
```
tcochran-quera, line 137: colored_nodes.add(node) # add this node and it's partner to the solved nodes.

src/bloqade/cirq_utils/noise/model.py
```diff
@@ -0,0 +1,542 @@
...
+from ..parallelize import parallelize
+from .conflict_graph import OneZoneConflictGraph
+
+
+@dataclass(frozen=True)
+class GeminiNoiseModelABC(cirq.NoiseModel, MoveNoiseModelABC):
+    """Abstract base class for all Gemini noise models."""
+
+    check_input_circuit: bool = True
+    """Determine whether or not to verify that the circuit only contains native gate.
```
tcochran-quera, line 19: """Determine whether or not to verify that the circuit only contains native gates.

src/bloqade/cirq_utils/noise/transform.py
```diff
@@ -0,0 +1,57 @@
...
+from ..parallelize import transpile, parallelize
+
+
+def transform_circuit(
+    circuit: cirq.Circuit,
+    to_native_gateset: bool = True,
+    model: cirq.NoiseModel | None = None,
+    parallelize_circuit: bool = False,
+) -> cirq.Circuit:
+    """Transform an input circuit into one with the native gateset with alternating 1Q and 2Q moments and add noise.
```
tcochran-quera, line 13: """Transform an input circuit into one with the native gateset with noise operations added.

src/bloqade/cirq_utils/noise/conflict_graph.py
```diff
@@ -0,0 +1,144 @@
...
+            deg_dict[node] = np.sum([node in edge for edge in self.edges])
+        self.degrees = deg_dict
+
+    def get_move_schedule(self, mover_limit: int = 10000):
+        """Generates a move schedule by coloring the conflict graph greedily, first coloring nodes of highest degree.
+
+        Qubits that are the arguments of a single CZ gate are 'partners'. Only one partner need be moved to arrange the
+        atoms for the 2Q gate. Thus, in coloring the conflict graph, as soon as one partner is colored, the other can be
+        disregarded for the purpose of coloring the rest of the graph.
+
```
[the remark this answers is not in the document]
david-pl: Actually, why is it setting an attribute here? It appears we only need the return value once.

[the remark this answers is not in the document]
tcochran-quera: That's right, in the current implementation, we don't need this attribute. Maybe in the future, we could add some other methods that act on `self.move_schedule`, but no plans for that at this stage.

weinbe58 (approved):

PR #370, opened 2025-07-08
Fix rewrite from squin -> stim with measurement

kaihsin:

PR #374, opened 2025-07-10
Allow slicing to work with address analysis

johnzl-777:
This allows slicing to work with the address analysis pass.

@liupengy19 ran into this problem with some of his work and @david-pl helped me confirm the issue was not specific to my setup (: 

Note: the way the tests are written for the address analysis are definitely a bit outdated, I believe once #370 goes through I updated the tests there so it will just be a matter of merging in the new updated stuff

### DIFF
--- a/test/analysis/address/test_analysis.py
+++ b/test/analysis/address/test_analysis.py
@@ -4,6 +4,7 @@ from kirin.passes import Fold
 from kirin.dialects import py, func, ilist
 
 from bloqade import qasm2, squin
+from bloqade.squin import qubit
 from bloqade.analysis import address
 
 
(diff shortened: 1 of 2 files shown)


test/analysis/address/test_analysis.py
```diff
@@ -203,6 +204,41 @@ def test_multiple_wire_apply():
...
+        if isinstance(address_qubit_type, address.AddressQubit)
+    ]
+
+    assert address_regs[0] == address.AddressReg(data=range(0, 4))
+    assert address_regs[1] == address.AddressReg(data=range(1, 3))
+
+    assert address_qubits[0] == address.AddressQubit(data=1)
+
+
+test_slice()
```
kaihsin, line 239: rm this

kaihsin (approved):

PR #384, opened 2025-07-18
Remove PPError and better typing for PauliError

### LINKED ISSUE #378: Merge squin's PPError and PauliError statements
They are quite similar and one is a subset of the other. The resulting statement should accept a `PauliString` as argument. Therefore, we should also rework the type system to have a `PauliStringType(OpType)`, which we use to restrict the argument type.

As potential syntax sugar, we could allow calling the wrapper directly with a string instead of a `PauliString` operator.

david-pl:
Closes #378

### DIFF
--- a/test/squin/cirq/test_cirq_to_squin.py
+++ b/test/squin/cirq/test_cirq_to_squin.py
@@ -174,6 +174,7 @@ def nested_circuit():
         swap_circuit,
         three_qubit_gates,
         noise_channels,
+        depolarizing_channels,
     ],
 )
 def test_circuit(circuit_f, run_sim: bool = False):


johnzl-777:
@david-pl I see your point, definitely a bit of a misunderstanding on my part on how lowering works because I haven't really had to deal with it until #381 

I suppose I was hoping there could be some way to have an `apply` that would be happy to accept a string and then we could do some heavy lifting when we see that string to inject the proper paulistring type. 

I suppose we'd have to go the current route of "try and quit" for such logic if we encounter a string that isn't pauli compatible, similar to how rewrites currently don't (intentionally) throw exceptions but rather just keep going and spit out a `RewriteResult(has_done_something=False)` 😅

david-pl:
@johnzl-777 Ah, yes, the misunderstanding is probably on my end.

You can just treat the first argument in an `apply` as the operator and if it's a string try to turn it into a `PauliString`. That works.

PR #392, opened 2025-07-22
Remove squin's broadcast

### LINKED ISSUE #350: Deprecate `squin.qubit.broadcast`?
The interface is unclear (weird strides in the list that depend on the qubit number) and overlaps with a for loop:
```python
controls:list[Qubit] = ...
targets:list[Qubit] = ...
op:Op = ...
for c,t in zip(controls,targets):
    squin.qubit.apply(op,c,t)

# equivalent to
register:list[Qubit] = interleave(controls,targets) #lol how do you do this actually? Where the evens are controls and odds are targets
squin.qubit.broadcast(op,register)
```

david-pl:
Closes #350.

As discussed, this PR removes squin's `broadcast` functionality since it's mostly the same as `apply`.

Also, the functionality of `apply` is extended so you can "broadcast" single-qubit operators, e.g.

```python
from bloqade import squin

@squin.kernel
def main():
    q = squin.qubit.new(3)
    x = squin.op.x()
    squin.qubit.apply(x, q)
    squin.qubit.apply(x, q[0], q[1], q[2])

main.print()
```

**NOTE:** we actually lose some functionality here as broadcasting for multi-site operators is no longer supported. While I tried to implement a syntax such as `squin.qubit.apply(cx, [[q[0], q[1], [q[2], q[3]])`, there were some issues that made me decide against supporting it altogether. Those issues are:

* We have to slap more and more functionality into the `ApplyDesugarRule`, which is bad enough as it is.
* Changing the type of the `Apply` statement such that it's argument is a list of lists breaks a lot of stuff.
* The main reason, however, was that you keep running into ambiguities. For example, you can't support `apply(cx, [q[0], q[1]])` and `apply(x, [q[0], q[1]])`, where in the latter I'd expect a broadcast. This is because in the rewrite you don't know the size of the operator. Both these calls are syntax I think we should have.
* We don't lose much (who is really using broadcast on multi-site operators?) and gain a lot in terms of simplicity. Also, we still support the application: just use a loop
```python
for i in range(n-1):
    squin.qubit.apply(cx, q[i], q[i+1])
```


david-pl:
@johnzl-777 

Sure, I can rename the stuff. Let's just clarify where we're going before doing that polishing work.

@weinbe58

> I know we can support unpacking tuples but to make sure we cover the use case @johnzl-777 pointed out this might be the way to go.

What use case would that be? It's not in the comment 😄 

Just so I get this right: You are proposing to leave `ApplyAny` as it is and simply change `Broadcast` to accept arbitrarily many lists of qubits. Why would we need any desugar then?

I'm going to challenge this a bit: I think the usefulness of being able to broadcast multi-site operators is limited. Right now (in this PR), you can broadcast single-site operators, which IMO is just a convenience functionality. The syntax of multi-site operator broadcasting is always a bit ambiguous, I think. And, quite frankly, we have the big advantage of actually being able to use control flow, so why not just write a loop? The meaning is always clear if you just index into the list.

Issue #347, opened 2025-07-01
Direct application of gates


jon-wurtz:
Instead of having to create a gate object and then apply it to qubits, have some boilerplate that directly applies it:

```python
qb = squin.qubit.new(1)[0]
h = squin.op.h()
squin.qubit.apply(h, qb)
```
is equivalent to
```python
qb = squin.qubit.new(1)[0]
squin.gate.h(qb)
```

Likewise equivalent
```python
qbs = squin.qubit.new(2)
cz = squin.op.cz()
squin.qubit.apply(h, qb[0], qb[1])
# is equivalent to
squin.gate.cz(qbs[0], qbs[1])
```

This is just sugar; in fact, these could be the "tinyest" kernels, eg
```python
@squin.kernel
def cz(ctrl:Qubit,target:Qubit):
    cz = squin.op.cz()
    squin.qubit.apply(cz,ctrl,target)
```
for every standard gate. Or, if there are parameterized gates its just
```python
@squin.kernel
def czphase(phase:float, ctrl:Qubit, target:Qubit):
    op = squin.op.czphase(phase=phase)
    squin.qubit.apply(op,ctrl,target)
```

david-pl:
Closed by #415

PR #393, opened 2025-07-22
RFC: direct application of simple squin operators

david-pl:
This came up in #347 

Basically, I think having the syntax

```python
q = squin.qubit.new(4)
squin.op.x(q[0])
```

would be nice to have, as it's short and intuitive.

I implemented a draft version in this PR so we can discuss the API. Having this is achieved by custom lowering, where we now support either zero or a single argument. So, we don't lose any previous functionality, but just add this on top. You can still write

```python
x = squin.op.x()  # --> Op
squin.qubit.apply(x, q[0])
```

This is completely equivalent to the shorthand above, and actually the shorthand gets lowered to the same IR.

## Open questions

1. The big one is, of course, do we even want this? But I'd argue the added complexity isn't too large and it's nice to have it.

2. I have only implemented this for operators that otherwise take zero arguments (e.g. `x(), y(), h(), ...`). It's convenient as you can easily share the same custom lowering here. However, my main reason for not doing that for any other operators is that I'm not sure what the API should look like. Take `u`, for example. It already takes three arguments, `u = squin.op.u(theta=pi/2, phi=0, lam=pi/4)`. How would we fit qubit(s) in there? Simply adding qubits as optional first or last argument seems clunky and it also makes the custom lowering somewhat more complex (and we'd potentially need separate custom lowering routines for different statements).
We could also just leave this unsupported, but I'm not sure. Ideas are very welcome!

3. One final small one: if we go this route, I'd even vote for importing the wrappers from `op` into `squin`, so you can even just write `squin.x(q[0])`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "bloqade-circuit"
-version = "0.6.0-DEV"
+version = "0.7.0-DEV"
 description = "The software development toolkit for neutral atom arrays."
 readme = "README.md"
 authors = [
--- a/src/bloqade/qasm2/_qasm_loading.py
+++ b/src/bloqade/qasm2/_qasm_loading.py
@@ -82,7 +82,7 @@ def loads(
         body=body,
     )
 
-    return ir.Method(
+    mt = ir.Method(
         mod=None,
         py_func=None,
         sym_name=kernel_name,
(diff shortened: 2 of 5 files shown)


david-pl:
Suggestion by @weinbe58:

Just add these as stdlib kernels, e.g. in a new submodule gate

```python
def x(q: Qubit):
    x = squin.op.x()
    squin.qubit.apply(x, q)

```

and a similar parallelization submodule that does `broadcast`.

PR #394, opened 2025-07-22
Fix nested for-loop handling in squin -> stim pass

johnzl-777:
Previously, the following would cause the squin to stim pass to fail:

```python
for i in range(...):
  for j in range(...):
    qubit.apply(op.cx(), [q[i], q[j])
```

This was because the unroll rule was not successful in fully unrolling the loops, getting (at most) one unroll before silently giving up. This behavior was first identified by @liupengy19 and encountered more recently by @ehua7365 as well.

@kaihsin pointed out to me that the problem might be because of a lack of constprop occuring between each unroll. This turns out to be the case and I've implemented a mini-pass that will do the constprop after each unroll, with the pass itself applied via fixpoint instead of the more traditional unsafe_run, single application style.

I've also taken the liberty to rename some of the tests to be a bit more accurate and removed some outdated comments.


cduck:
Why make a new pass not fix the old unroll pass?

johnzl-777:
@cduck @david-pl I can confirm there was an even more minimal MWE that just used `@structural` from kirin's `prelude`.

I can make an issue on kirin for this but my inclination to fix it here is because I know it takes time for kirin fixes to be propagated. I don't recall @liupengy19 mentioning this was particularly blocking but you can tell the lack of such an ability makes for poor ergonomics

david-pl:
@johnzl-777 I see, this makes sense and the PR looks good to me. But please create an issue on kirin and reference this PR. Also, it might make sense to create a follow-up issue in bloqade-circuit so we don't forget to "undo" this change here once we have the proper fix in kirin and are at the new version.

david-pl (approved):

PR #405, opened 2025-07-24
fix nested list handling in SquinToStimPass

### LINKED ISSUE #404: StimSimplifyIfs fails on nested lists
@liupengy19 alerted me to a regression caused by https://github.com/QuEraComputing/bloqade-circuit/pull/394

Previously, something like:
```
    pairs=[[0,1],[2,3]]
    @squin.kernel
    def main():
        q = qubit.new(10)
        h=op.h()
        for i in range(2):
            qubit.apply(h,q[pairs[i][0]])
```

Was able to be properly converted to stim. Now this seems to fail. Part of this is attributable to the fact that nested lists are not fully lowered into (that is, not all elements become) ILists.

More details and the solution I found is detailed below from the original discussion.

> 
> Basically in both pre AggressiveUnroll and post aggressiveUnroll, the behavior that @cduck mentions is present. The members of the nested list do NOT get lowered into ILists as well. 
> 
> I would have expected this gum things up much earlier but instead, it seems the only reason it is a problem NOW is there is too _much_ information available to the compiler and in its eagerness to optimize things it gets into this problem. 
> 
> The specifics look like the following:
> 1. HintConst is explicitly invoked after every application of Unroll which gives subsequent Unroll calls the data they need by checking for the "const" hint in the SSA Value
> 2. At the tail end of the SimplifyIfsToStim pass, after the scf.IfElse have been massaged into a form that can be rewritten to feedforward ops (ex: CX rec[-1] 0) there is some additional logic to clean things up
> 3. The cleaning involves repeatedly calling ConstantFold and CSE. ConstantFold sees that there is the "const" hint and begins to try to pull the internal elements of the nested list. Thus, you get a bunch of 'consts' added that are just pure python lists
> 4. CSE attempts to hash that pure python list and immediately grinds to a halt because lists are not hashable
> 
> The solution is that instead of something like ConstantFold -> CSE, you need ConstantFold -> ConstList2IList -> THEN you can safely do CSE (: 
>

johnzl-777:
Fixes #404 and also caught a mistake I introduced which is that AggressiveUnroll does not `join` the rewrite_result chain.

### DIFF
--- a/test/stim/passes/test_squin_qubit_to_stim.py
+++ b/test/stim/passes/test_squin_qubit_to_stim.py
@@ -190,6 +190,13 @@ def test_nested_for_loop_rewrite():
 
 def test_nested_list():
 
+    # NOTE: While SquinToStim now has the ability to handle
+    # the nested list outside of the kernel in this test,
+    # in general it will be necessary to explicitly
+    # annotate it as an IList so type inference can work
+    # properly. Otherwise its global, mutable nature means
+    # we cannot assume a static type.
+
     pairs = [[0, 1], [2, 3]]
 
     @squin.kernel


kaihsin (approved):

david-pl:
@johnzl-777 looks good to me, but CI appears to be failing.
Also, I again have the question whether this is actually an upstream issue? I feel like this should be lowered properly in kirin.

david-pl (changes requested):
Please fix the CI

kaihsin:
> @johnzl-777 looks good to me, but CI appears to be failing. Also, I again have the question whether this is actually an upstream issue? I feel like this should be lowered properly in kirin.

I am ok with temporary fix here, but please create an MWE and open an issue on kirin. ty!

kaihsin:
This has been an issue for nested IList on kirin side. I don't think the RFC is resolved there

kaihsin:
https://github.com/QuEraComputing/kirin/issues/277

PR #412, opened 2025-07-26
add alias support to measure_id analysis

johnzl-777:
@ehua7365 encountered a problem where reassigning an IList of measure results would cause `SquinToStim` to fail:

```
        q = qubit.new(4)
        ms = qubit.measure(q)
        new_ms = ms

        if new_ms[0]:
            qubit.apply(op.z(), q[0])

```
(This is an intentionally contrived example to keep things short). Originally I had directed Eric to include kirin's InlineAlias and InlineGetItem rules which also got rid of the issue until I took a closer look and realized the problem is because when an alias happens with an IList of measure results, the analysis does not propagate the corresponding data. 

As a result, further Stim-related rewrites have problems because the MeasureIDAnalysis data is not available for the subsequent SSA Values.


kaihsin:
@johnzl-777 why is InlineAlias does not remove that alias?

johnzl-777:
@kaihsin InlineAlias still works! If that's more preferable I can just do that instead but I thought this would be a bit more preferable

PR #413, opened 2025-07-26
add ability to track number of measurements at statements of interest

johnzl-777:
Every time an scf.IfElse is encountered the total number of measurements that have occurred will be saved and accessible in later rewrites


kaihsin:
LGTM, maybe just add more tests. The cov is failing @johnzl-777

Issue #435, opened 2025-08-05
RFC: tracking new squin statements


david-pl:
We repeatedly run into issues where a new squin statement is added to the dialect without adding all the corresponding implementations in all method tables / rewrites. From a user perspective, I'd expect that any statement available in a dialect works in all places.

@cduck proposed to add a unit test. For example, we could have a kernel that needs to include every single dialect statement (checked by looping over all available statements and checking). Then, we can run this kernel through all required implementations and check the result.
The only disadvantage of this is that it may block PRs that add new statements. Then again, I think this is the only sure-fire way to avoid missing implementations.

We'll also need a list of all the places we need to run the kernel through. Right now, I can think of the following:

* Noise to `StochasticUnitary` rewrite rule
* PyQrack simulation
* cirq emit
* `SquinToStimPass`
* Address analysis

@johnzl-777 @kaihsin @weinbe58 any other places? I suppose there are some things that need to be tested downstream as well, but this can't be included here.

**Conclusion**:

With the refactor that happened in v0.8, squin has been somewhat simplified and the number of statements has been greatly reduced. We shouldn't add new statements frequently, but instead wrap them in a standard library function. That way, this issue should no longer be relevant.

johnzl-777:
> [@cduck](https://github.com/cduck) proposed to add a unit test. For example, we could have a kernel that needs to include every single dialect statement (checked by looping over all available statements and checking). 

Just to be precise here, you mean that the kernel would be created by hand but then there would be a way to check that kernel contains every possible squin statement, right?


> Then, we can run this kernel through all required implementations and check the result. The only disadvantage of this is that it may block PRs that add new statements. Then again, I think this is the only sure-fire way to avoid missing implementations.

I think it's probably a good thing to have some (more) "mechanical"/"code" level friction on top of the human element (RFCs and triage if things become contested) (: 

If we're forced to add logic to method tables/rewrites for new statements we can find out sooner rather than later if it proves to be particularly clunky/not worth it.


> We'll also need a list of all the places we need to run the kernel through. Right now, I can think of the following:
> 
>     * Noise to `StochasticUnitary` rewrite rule
> 
>     * PyQrack simulation
> 
>     * cirq emit
> 
>     * `SquinToStimPass`
> 
>     * Address analysis
> 
> 
[...]

PR #436, opened 2025-08-05
Test basic methods of all squin statements

david-pl:
Draft for #435 

Looks like this was long overdue. It fails currently because:

* Missing noise rewrite for `Depolarize2` (this also means pyqrack simulation fails)
* Cirq emit is missing `Rot` (wow, that's on me) and `ResetToOne`.
* There's a bug in the `Squin2StimPass` which causes an error.

The only thing that seems to work right now is the `AddressAnalysis` (kudos @johnzl-777).

@kaihsin @johnzl-777 the test for the `Squin2StimPass` still needs some work to properly test stuff, but I didn't want to mix commits and start fixing bugs right now, so we need to add that later.
FYI, I added a little rewrite pass just for testing that removes intentionally unsupported statements in case that's relevant for the stim tests. We should probably document these statements somewhere, but right now it's basically just atom loss that is unsupported in cirq.

**Open questions**:
* Is this everything or did I miss out on something?
* How do we test the wire dialect?

### DIFF
--- a/test/squin/test_all_statements.py
+++ b/test/squin/test_all_statements.py
@@ -1,0 +1,291 @@
+from types import ModuleType
+from dataclasses import dataclass
+
+from kirin.ir import Statement
+from kirin.rewrite import Walk
+from kirin.rewrite.abc import RewriteRule, RewriteResult
+
+from bloqade import squin
+from bloqade.pyqrack import StackMemorySimulator
+from bloqade.analysis import address
+from bloqade.stim.passes import SquinToStimPass
+from bloqade.squin.noise.rewrite import RewriteNoiseStmts
+
+
+def get_dialect_stmts(module, stmts_module):
+    members = [getattr(stmts_module, stmt) for stmt in dir(stmts_module)]
+
+    stmts = set()
+    module_dialect = getattr(module, "dialect")
+    for member in members:
+        try:
+            dialect = member.dialect
+        except AttributeError:
+            # not a stmt
+            continue
+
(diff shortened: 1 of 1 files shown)


david-pl:
@johnzl-777 not sure what the problem with the `Squin2Stim` rewrite is right now. It chokes when I try to add the auxiliary dialect (which I probably shouldn't be doing in any case). Is there a key clash somewhere?

david-pl:
Alright, looks like the squin2stim test is the last one we need to fix, all other tests pass already.

johnzl-777:
@david-pl I'm still not quite sure why `stim.aux` was causing so much trouble but I suspect it has something to do with the combination of fixpoint being called instead of the standard unsafe_run and the type inference pass. 

That being said just a couple observations on my end:
- SquinToStim CAN handle specific instances of scf.IfElse and U3 (equivalent to Clifford gates). I can modify your original kernel slightly to allow this to happen
- There's currently no "adjoint" canonicalization that happens right now (that is, something clever where if you have nested adjoints you can collapse them) and while we do have the ability to emit things like SQRT_X_DAG it turns out there's some incorrect logic in SquinToStim I'll have to fix (I literally just realized from trying to get this unit PR to work haha)
- is it desirable to have these tests assert against something? I understand these tests at the very least force relevant passes/analyses targeting squin to run but I feel like it should be checked against something. If it's desirable I can go ahead and just do what I've been doing with the SquinToStim tests which is assert against a `.stim` file.

PR #438, opened 2025-08-05
Support Slice propogation in MeasurementIDAnalysis

johnzl-777:
I was working on a more advanced unit test for SquinToStim when I ran into this problem. 

If you use slice on a list of measurements the analysis doesn't go through. The solution was to just carry over the logic I figured out for the address analysis. However I cooked up a case where even this is insufficient and figured out the simplest addition to the logic to amend this case.

Basically both measure and address have something like the following:
```
# container could be AddressTuple or MeasureIDTuple
getitem(container, slice) -> tuple(....)
```
Where tuple is literally a plain Python tuple. This doesn't present any problems until you slice again:

```
getitem(tuple, slice) -> ???
```
in which case the method table impl is blind to how to handle a raw tuple. I believe the more correct way is that slice should generate ANOTHER container type instead of a naked python tuple.


johnzl-777:
I think this logic should be added to address analysis as well to make it more robust if you agree with the implementation here (:

johnzl-777:
@weinbe58 my understanding is it doesn't work for cases like output from `qubit.new` or `qubit.measure`, which is what I'm trying to target here. InlineGetItem is very nifty in SquinToStim though (:

src/bloqade/analysis/measure_id/analysis.py
```diff
@@ -37,20 +37,17 @@ def run_method(self, method: ir.Method, args: tuple[MeasureId, ...]):
...
     # Xiu-zhe (Roger) Luo came up with this in the address analysis,
-    # reused here for convenience
+    # reused here for convenience (now modified to be a bit more graceful)
     # TODO: Remove this function once upgrade to kirin 0.18 happens,
     #       method is built-in to interpreter then
-    def get_const_value(self, input_type: type[T], value: ir.SSAValue) -> T:
+
+    T = TypeVar("T")
+
+    def get_const_value(self, input_type: T, value: ir.SSAValue) -> T | None:
```
kaihsin, line 47: it should be `type[T]` not `T` for annotation for `input_type`!

kaihsin: as well as the return type

johnzl-777: My bad, this was left over from me wrestling with Pylance, I was trying to get it to be happy with the arguments I was passing in

PR #441, opened 2025-08-06
Implement squin noise rewrite for depolarize2

david-pl:
Another thing that came out of #436.


david-pl:
The second commit actually fixes the bug from #446 .

weinbe58 (approved):
LGTM

PR #442, opened 2025-08-06
Add PauliString impl to NsitesAnalysis

david-pl:
Yet another bug found in #436

### DIFF
--- a/src/bloqade/squin/analysis/nsites/impls.py
+++ b/src/bloqade/squin/analysis/nsites/impls.py
@@ -81,6 +81,13 @@ class SquinOp(interp.MethodTable):
         op_sites = frame.get(stmt.op)
         return (op_sites,)
 
+    @interp.impl(op.stmts.PauliString)
+    def pauli_string(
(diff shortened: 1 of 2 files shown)


src/bloqade/squin/analysis/nsites/impls.py
```diff
@@ -81,6 +81,13 @@ def scale(self, interp: NSitesAnalysis, frame: interp.Frame, stmt: op.stmts.Scal
         op_sites = frame.get(stmt.op)
         return (op_sites,)
 
+    @interp.impl(op.stmts.PauliString)
+    def pauli_string(
+        self, interp: NSitesAnalysis, frame: interp.Frame, stmt: op.stmts.PauliString
+    ):
+        s = stmt.string
+        return (NumberSites(sites=len(s)),)
```
johnzl-777, line 89: I think it would be a bit more robust if when `len(s)==0` you return `NoSites()` (:

johnzl-777 (changes requested):

PR #444, opened 2025-08-06
Reduced density matrix to pyqrack

jon-wurtz:


jon-wurtz:
As part of the addition, I removed the `state_vector` method, as the actual state vector has unclear basis

david-pl:
@jon-wurtz thanks! I think the implementation looks quite solid, but I have some comments regarding the API.

* Do I understand correctly that this returns a state vector of the given qubits if the resulting state is pure, and a matrix otherwise? If so, calling this "reduced density matrix" might be a bit confusing since it's not always a matrix.
* Do we actually need to return the eigenvalues? Or can we at least make it optional with a default to not return them? I feel like you'd just expect to get a state as a return here.
* Please add `ilist.IList[PyQrackQubit, Any]` as a possible argument to the method, since that's what you're actually calling the method with and the linter is complaining.
* Maybe we should also make it possible to call the method with a single qubit.
* I also think it would be natural to write something like `emulator.reduced_density_matrix([0,1,2])` (or with respect to the point just above `emulator.reduced_density_matrix(0)`), where the integers are the indexes of the corresponding register (not the Qrack qubit address). It would also allow you to obtain the states without returning the list of qubits from the kernel without having to build a `list[PyQrackQubit]` manually, which is a bit clunky.

david-pl:
> @david-pl changes have been addressed, please review.

Well, the code is practically the same, besides being moved around 😅 

> I'm going to stay stalwart on keeping consistent input and output signatures.

Okay, fair enough. However, in terms of UX I'd still suggest to change things a little here. I understand that you return a faithful representation of the reduced density matrix. Still, I guarantee that this will trip up users, especially new ones. If I call a method named `reduced_density_matrix` I expect exactly that.

My suggestion: rename the method you have (not entirely sure to what, maybe `reduced_density_matrix_eigh` or even `reduced_state`?). Then, create a convenience method that calls this method and does the "busy work" - as you call it - for the user and returns the actual matrix. That method would then be called `reduced_density_matrix`.

This would have consistent APIs and return value signatures and allow the user to get both the reduced density matrix or the more efficient representation depending on their use case. Everybody wins ;)

src/bloqade/pyqrack/device.py
```diff
@@ -112,6 +162,51 @@ def pauli_expectation(pauli: list[Pauli], qubits: list[PyQrackQubit]) -> float:
 
         return sim_reg.pauli_expectation(qubit_ids, pauli)
 
+    @staticmethod
+    def reduced_density_matrix_rdm(
```
david-pl, line 166: Very much optional, but this now just reads as "reduced-density-matrix-reduced-density-matrix" in my mind, so I'm not sure about the naming 😅

src/bloqade/pyqrack/device.py
```diff
@@ -112,6 +162,51 @@ def pauli_expectation(pauli: list[Pauli], qubits: list[PyQrackQubit]) -> float:
...
+    ) -> np.ndarray:
+        """
+        Extract the reduced density matrix representing the state of a list
+        of qubits from a PyQRack simulator register.
+
+        Inputs:
+            qubits: A list of PyQRack qubits to extract the reduced density matrix for
+            tol: The tolerance for density matrix eigenvalues to be considered non-zero.
+        Outputs:
+            An eigh result containing the eigenvalues and eigenvectors of the reduced density matrix.
```
david-pl, line 203: That doc string is now outdated. Please update.

src/bloqade/pyqrack/task.py
```diff
@@ -36,3 +37,12 @@ def state_vector(self) -> list[complex]:
         """Returns the state vector of the simulator."""
         self.run()
         return self.state.sim_reg.out_ket()
+
+    def qubits(self) -> list[PyQrackQubit]:
+        """Returns the qubits in the simulator."""
+
+        N = self.state.sim_reg.num_qubits()
```
david-pl, line 44: If you don't run the task before calling this method, you'll get a not very helpful error message: ´AttributeError: 'StackMemory' object has no attribute 'sim_reg'´.

I'd suggest you wrap the access to ´.sim_reg´ in a ´try - except´ (gotta love idiomatic python) and run the task if you catch an ´AttributeError´.

Also, it would be nice to have this method available directly on the simulator, similar to `.state_vector`.

david-pl (changes requested):
@jon-wurtz thanks, this is looking very good now. Just some minor comments and then I think this is good to go!

david-pl (approved):

Issue #448, opened 2025-08-08
Remove uncorrelated noise channel?


tcochran-quera:
I don't know if there is actually any reason to keep the [uncorrelated noise model](https://github.com/QuEraComputing/bloqade-circuit/blob/339e60d8c48dc5640ec104b61527921eaf731025/src/bloqade/cirq_utils/noise/model.py#L140), as I think the [correlated noise model](https://github.com/QuEraComputing/bloqade-circuit/blob/339e60d8c48dc5640ec104b61527921eaf731025/src/bloqade/cirq_utils/noise/model.py#L304) should be the default tool.

tcochran-quera:
@david-pl thoughts?

david-pl:
No objections from my side, although it will technically be a breaking change (probably not too many people use it though). We'll have to update the example in the docs accordingly.

The only thing that comes to mind is whether we want to have a nicer way of setting the correlated noise parameters.

cirq's dictionary syntax for multiple qubit depolarization comes to mind. They have something like `cirq.asymmetric_depolarize(error_probabilities={'XX': 0.1, 'XY': 0.01, 'YX': 0.01})`, where any key missing from the dictionary is just zero.
This could potentially be done with a helper function / optional argument to the noise model, although I'm not sure if you don't always need a full array anyway.

Just to illustrate what I mean, we could have something like

```python
cz_paired_noise: np.ndarray = get_cz_paired_noise(error_probabilities = {'00': 1e-3, '01': 1e-4, ...}
```

or, equivalently

```python
from bloqade.cirq_utils import noise
noise_model = noise.GeminiOneZoneNoiseModel(cz_paired_noise = {'00': ...})
```

Not sure if we really need this as I don't know how often users will actually change the defaults of this particular noise process.

Also, are there any constraints on this matrix other than that it needs to be a real and symmetric 4x4 matrix? We should probably check the input here.

PR #459, opened 2025-08-12
Make all GeminiOneZoneModels use correlated noise

david-pl:
Implements #448.

The test on the noisy GHZ fidelity now fails. On the one hand, of course, since the fidelities did change. On the other hand, however, we now somehow get a singular matrix for two qubits which produces a corresponding `np.LinAlg` warning and results in a fidelity > 1. FWIW, checking manually you can see that the matrices are pretty close, so the issue really seems to be with the way the fidelity is calculated.

What's also kind of fishy: the fidelity for three qubits increased (slightly). Before, we had: 0.9518, now it's 0.957. @tcochran-quera @lamq317 did you see something similar when comparing the one-zone model with and without correlated noise or is there something wrong here?

You can now also pass in the correlated rates either as dictionary or as array

```python
model = GeminiOneZoneModel(cz_paired_correlated_rates = np.array(...))
model = GeminiOneZoneModel(cz_paired_error_probabilities = {"II": 0.9, "XX": ...})
```

Not sure about the naming though.

### DIFF
--- a/src/bloqade/cirq_utils/noise/_two_zone_utils.py
+++ b/src/bloqade/cirq_utils/noise/_two_zone_utils.py
@@ -266,7 +266,7 @@ def get_gate_error_channel(
     sq_glob_rates: np.ndarray,
     two_qubit_pauli: cirq.Gate,
     unp_cz_rates: np.ndarray,
-    nqubs
+    nqubs: int,
 ):
     """Applies gate errors to the circuit
 
@@ -283,7 +283,6 @@ def get_gate_error_channel(
     # Check for the moment (layer) layout: global single qubit gates, or mixture of single qubit gates and two qubit gates
 
     gates_in_layer = extract_u3_and_cz_qargs(moment)
-    # new_moment = cirq.Moment()
     new_moments = cirq.Circuit()
 
     if gates_in_layer["cz"] == []:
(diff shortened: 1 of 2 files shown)


src/bloqade/cirq_utils/noise/model.py
```diff
@@ -81,72 +149,33 @@ def cz_unpaired_pauli_rates(self) -> tuple[float, float, float]:
             self.cz_unpaired_gate_pz,
         )
 
+    @property
+    def two_qubit_pauli(self) -> cirq.AsymmetricDepolarizingChannel:
+        # NOTE: if this was None it would error when instantiating self
+        # need to make the linter shut up about the copy below
```
tcochran-quera, line 155:
```suggestion
        # quiet the linter for the copy below
```

tcochran-quera (approved):
LGTM with one nit, but maybe you were planning on removing that NOTE anyway?

I did notice that printing the new annotated circuits is pretty unruly because it prints the entire 2Q asymmetric 2Q dictionary. I guess it was always this way and I'm just noticing it more now since we're migrating to the correlated noise channel being the default.  No changing the `_circuit_diagram_info` in cirq.

As a longer term project, I wonder if it makese sense to put together a interactive circuit that just shows that gates, but when one hovers the mouse over a particular gate, all the associated move errors are displayed. What do you think?

image

david-pl:
Thanks for the quick review @tcochran-quera !

> As a longer term project, I wonder if it makese sense to put together a interactive circuit that just shows that gates, but when one hovers the mouse over a particular gate, all the associated move errors are displayed. What do you think?

This sounds very cool, but we currently don't have any circuit display methods, so it's definitely some time away. Maybe this should be raised with cirq instead since it's their printing we're talking about? Not sure if they have any other fancy display methods though.

PR #461, opened 2025-08-13
Implement PhasedXZ gate statement in squin

david-pl:
@weinbe58 After working on this I'm no longer so sure if we actually want to support it since it is a bit of a hassle (see for example the pyqrack method impl).

I also added a rewrite to decompose the statement into rotations, which I then decided not to use. But I already had it so here it is.


weinbe58:
we can close this for the time being, if we have any issues later on we can add this back

PR #469, opened 2025-08-18
Use CallGraphPass to rewrite noise stmts

david-pl:
This is needed in order for simulations with PyQrack to "just work" when using the new stdblib short-hands such as `squin.channel.depolarize(p=0.1, qubit=q[0])`, since those are `Invoke` statements.

@weinbe58 I'm 100% about the API of `CallGraphPass` here. The options I see are

**Option 1**
Inherit from `CallGraphPass` and hardcode the `rule` attribute in order to have something that is basically just a `Pass`, but does its thing on the entire callgraph. This is what I committed.
Feels a bit weird to just override the attribute here.

**Option 2**
Define a function

```python
def RewriteNoiseStmts(dialects, **kwargs):
    rule = Walk(_RewriteNoiseStmts())
    return CallGraphPass(dialects=dialects, rule=rule, **kwargs)
```

which makes this not a `Pass` object.

Let me know what you think!

### DIFF
--- a/src/bloqade/qasm2/rewrite/register.py
+++ b/src/bloqade/qasm2/rewrite/register.py
@@ -2,7 +2,7 @@ from kirin import ir
 from kirin.dialects import py
 from kirin.rewrite.abc import RewriteRule, RewriteResult
 
-from bloqade.qasm2.dialects import core
+from bloqade.qasm2.dialects import core, expr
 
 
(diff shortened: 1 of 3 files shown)


david-pl:
This is pending on the decision of the RFC #476. The implementation of the RFC in #478 might make this PR obsolete.

PR #472, opened 2025-08-18
Clearly document the units of angles in wrappers

### LINKED ISSUE #329: make units more explicit for rotation wrappers
we find a lot of people constantly get confused about the units when using these gates, so one proposal we come up during standup is add a suffix to these gates, e.g `U3` -> `U3rad` I'm wondering if there are other cases we should add a suffix? e.g `rot` maybe?

david-pl:
Closes #329 .

@Roger-luo I'm wondering whether this is sufficient. There was no documentation before so maybe that was just the issue. I don't know, I just really dislike the idea of having e.g. `rot_rad`. It also doesn't make complete sense since there is another argument there that's not an angle. I will change to that if you think just having nice docstrings is insufficient.

### DIFF
--- a/src/bloqade/squin/op/_wrapper.py
+++ b/src/bloqade/squin/op/_wrapper.py
@@ -50,15 +50,33 @@ def identity(*, sites: int) -> types.Op: ...
 
 
 @wraps(stmts.Rot)
-def rot(axis: types.Op, angle: float) -> types.Op: ...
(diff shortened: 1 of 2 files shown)


Roger-luo:
I think this came out from a meeting discussion at some point, unfortunately I don't remember the original issuer since I put it. I'm personally ok with not having the postfix, or just let the argument name taking a `_rad` postfix so when the inlay hints are turned on, people who cares can still see it.

Issue #468, opened 2025-08-15
Smarter Adjoint Handling in SquinToStimPass (and in general)


johnzl-777:
I ran into this problem debugging the `SquinToStimPass` in @david-pl 's ongoing PR #436 which uses all squin statements against relevant passes/analyses to ensure they're all accounted for. 

I suppose this RFC is a combination of asking what the best way to resolve this is and pushing for something I thought of some time ago but feel wouldn't hurt to have now.

The `Adjoint` statement in squin has a boolean attribute that determines if it's unitary. By default, it's set to `False`. However, there has to be some logic to be able to set it to `True` when the operator inside the adjoint is unitary. The motivating example for this is currently if I have something like:

```
op = adjoint(sqrt_x)
```
This will not get properly expressed as SQRT_X_DAG in the rewrite to Stim. The reason is because there is some guard logic I believe @kaihsin implemented where if the adjoint is not unitary, then the squin qubit dialect to stim rewrite immediately stops.

I can see the following solutions to the problem:
1. Just make the guard a bit more relaxed and check if what's inside the adjoint is a SqrtX, SqrtY, or SqrtZ to allow things to go through
2. Have a rewrite that looks at instances of such adjoints and fixes up the boolean to reflect the unitarity

This brings me to another idea that I don't think is too much work. This is more directly related to what @david-pl has in his test where he has `adjoint(x)` (x being Pauli X). There's currently no adjoint canonicalization pass but I think it would quite trivial to have some rewrite that can see see such things and realize the adjoint can be removed altogether or a chain of adjoints can be collapsed. 

Please feel free to let me know if I should take the simplest route here to make the SquinToStimPass happy or if it's worth killing two birds with one stone 🪨 🐦 🐦

kaihsin:
Canonicalize pass would be better

PR #475, opened 2025-08-19
Implement hermitian and unitary analysis and canonicalization pass

david-pl:
Still needs better testing, but works in principle.

I'm abusing `EmptyLattice` here a bit. Technically, I think you'd need a lattice to represent

* `Hermitian`
* `NotHermitian`
* `NotAnOperator`

and similar for unitary.
I'm just doing this now with `True`, `False` and `missing`.

### DIFF
--- a/src/bloqade/squin/analysis/hermitian_and_unitary.py
+++ b/src/bloqade/squin/analysis/hermitian_and_unitary.py
@@ -1,0 +1,173 @@
+from kirin import ir, types, interp
+from kirin.interp import InterpreterError
+from kirin.lattice import EmptyLattice
+from kirin.analysis import Forward, ForwardFrame, const
+from kirin.dialects import scf, func
+
+from .. import op
+
+
+class HermitianAnalysis(Forward):
+    keys = ["squin.hermitian"]
+    lattice = EmptyLattice
+
+    def run_method(self, method: ir.Method, args: tuple[EmptyLattice, ...]):
+        return self.run_callable(method.code, (self.lattice.bottom(),) + args)
+
+    def eval_stmt_fallback(
+        self, frame: ForwardFrame, stmt: ir.Statement
+    ) -> tuple[bool] | None:
+        if not isinstance(stmt, op.stmts.Operator):
+            return
+
+        if stmt.has_trait(op.traits.Hermitian):
+            return (True,)
+
+        if (trait := stmt.get_trait(op.traits.MaybeHermitian)) is not None:
+            return (trait.is_hermitian(stmt),)
+
+        return (False,)
+
+
+class UnitaryAnalysis(Forward):
+    keys = ["squin.unitary"]
+    lattice = EmptyLattice
+    hermitian_frame: ForwardFrame | None = None
+
+    def run_method(self, method: ir.Method, args: tuple[EmptyLattice, ...]):
+        hermitian_frame, _ = HermitianAnalysis(method.dialects).run_analysis(method)
+
+        # NOTE: we need to update this if it's already set, otherwise an invoke will overwrite the frame
+        if self.hermitian_frame is None:
+            self.hermitian_frame = hermitian_frame
+        else:
+            self.hermitian_frame.entries.update(hermitian_frame.entries)
+
+        return self.run_callable(method.code, (self.lattice.bottom(),) + args)
+
+    def eval_stmt_fallback(
+        self, frame: ForwardFrame, stmt: ir.Statement
+    ) -> tuple[bool] | None:
+        if not isinstance(stmt, op.stmts.Operator):
+            return
+
+        if stmt.has_trait(op.traits.Unitary):
+            return (True,)
+
+        if (trait := stmt.get_trait(op.traits.MaybeUnitary)) is not None:
+            return (trait.is_unitary(stmt),)
+
+        return (False,)
+
+
+@op.dialect.register(key="squin.hermitian")
+class HermitianMethods(interp.MethodTable):
+    @interp.impl(op.stmts.Control)
+    @interp.impl(op.stmts.Adjoint)
+    def simple_container(
+        self,
+        interp: HermitianAnalysis,
+        frame: ForwardFrame,
+        stmt: op.stmts.Control | op.stmts.Adjoint,
+    ):
+        return (frame.get(stmt.op),)
+
+    @interp.impl(op.stmts.Scale)
+    def scale(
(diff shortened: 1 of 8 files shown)


kaihsin (changes requested):
I would suggest to just make proper lattice not abusing EmptyLattice. For future extension

src/bloqade/squin/analysis/hermitian_and_unitary.py
```diff
@@ -0,0 +1,161 @@
...
+        interp: UnitaryAnalysis,
+        frame: ForwardFrame,
+        stmt: op.stmts.Control | op.stmts.Adjoint,
+    ):
+        return (frame.get(stmt.op),)
+
+    @interp.impl(op.stmts.Scale)
+    def scale(self, interp: UnitaryAnalysis, frame: ForwardFrame, stmt: op.stmts.Scale):
+        if not frame.get(stmt.op):
+            # NOTE: definitely not hermitian
```
johnzl-777, line 112:
```suggestion
            # NOTE: definitely not unitary
```

src/bloqade/squin/op/stmts.py
```diff
@@ -87,15 +109,18 @@ class Control(CompositeOp):
 
 @statement(dialect=dialect)
 class Rot(CompositeOp):
```
johnzl-777, line 111: Is there a reason `Rot` doesn't have any hermitian trait?

johnzl-777 (changes requested):
Looks good! +1 on @kaihsin 's feedback. Just to plan ahead a bit, with the stuff I brought up in #468 the goal is not to knock that out here, right? Although this infrastructure is certainly necessary to do so!

src/bloqade/squin/op/stmts.py
david-pl: No not really. I can add it, but it's only hermitian if the angle is 0, in which case it's actually just the identity. So the use-case is just very limited, which is why I skipped it.

david-pl: Oh, this also applies to `U3`, `PhaseOp` and `ShiftOp` (maybe a few more). I feel like it's a bit of an overkill to have the hermitian analysis in place here, as it's always only the case that they are hermitian when the exponent is 0 and therefore the operator is just the identity. What do you think @johnzl-777 ?

As a side note: I generally think this should be a rewrite pass, which just replaces the operator by the identity directly. But we don't really have any such simplification passes in-place right now. For example, simplifying multiplications with identity operators would be another such candidate.

johnzl-777:
> and similar for `Unitary`. I'm not sure if have `NotHermitian` as top is 100% correct in terms of semantics, but in the implementations you always reach that top when you can't decide whether an operator is hermitian, so treating it as `NotHermitian` is fine, I think. The rewrite pass that sets the attribute would behave exactly the same on `NotHermitian` and another top such as `AnyHermitian`.
> 

My understanding of the top element in the type lattice is there is the possibility it COULD be any other lattice element but there's insufficient information to narrow it down. In that case I think something like `PossibleHermitian` might be better here as well as `PossibleUnitary`

src/bloqade/squin/analysis/unitary/impls.py
```diff
@@ -0,0 +1,66 @@
...
+    def scale(self, interp: UnitaryAnalysis, frame: ForwardFrame, stmt: op.stmts.Scale):
+        is_unitary = frame.get(stmt.op)
+
+        # NOTE: need to check if the factor has absolute value squared of 1
+        constant_value = stmt.factor.hints.get("const")
+        if not isinstance(constant_value, const.Value):
+            return (NotUnitary(),)
+
+        num = constant_value.data
+        if not isinstance(num, (float, int, bool, complex)):
```
johnzl-777, line 33: Huh, just learned that `abs` works on bools but now I have to ask, is there actually a case where somebody will pass in a boolean for the `factor` arg considering we limit it to `NumberType`?

src/bloqade/squin/op/stmts.py
johnzl-777: I see, I think your proposal for a rewrite makes sense here instead of burdening other statements with the hermitian information. I suppose I'm just nervous how much of a downside we could get if it turns out something IS hermitian (this 0 exponent case) and we don't catch it. 

I imagine it wouldn't be the worst thing in the world and if we knew we'd have something spitting out 0-exponents we'd most likely have thought to include that rewrite in the first place.

src/bloqade/squin/rewrite/canonicalize.py
```diff
@@ -58,3 +62,68 @@ def rewrite_Statement(self, node: ir.Statement) -> abc.RewriteResult:
...
+            return abc.RewriteResult()
+
+        is_unitary = self.unitary_values.get(node.result)
+        if is_unitary is None:
+            return abc.RewriteResult()
+
+        if is_unitary.is_equal(Unitary()):
+            maybe_unitary.set_unitary(node, True)
+        else:
+            maybe_unitary.set_unitary(node, False)
```
johnzl-777, line 112:
```suggestion
        new_unitary_status = is_unitary.is_equal(Unitary())
        maybe_unitary.set_unitary(node, new_unitary_status)
```
Just feels a little tidier to me

johnzl-777 (changes requested):
I'm still a bit hesitant to have a lattice top element have "Not" as part of its name and I also think it a bit odd to have a check for `bool` surrounding the `scale` impl (even though I just now learned it's literally subclassed from `int`)

david-pl:
> I'm still a bit hesitant to have a lattice top element have "Not" as part of its name

@johnzl-777 I agree. While working out the solution, however, I noticed a problem, which I now realize is why I wrote things like this initially:

Any basic operator statement that is not hermitian or not unitary represents this by an absence of the respective trait. For example, `SqrtX` is unitary, but not hermitian. So in that sense, they are not really distinguishable from an operator for which we can't infer whether it is hermitian or not.

I solved this by adding a simple check for `PrimitiveOp` in the analysis fallback, which should be fine, I suppose, but it might miss a few things.

src/bloqade/squin/analysis/unitary/impls.py
johnzl-777: Right, I admit I'm splitting hairs here I just think the presence of `bool` itself here implies we support it in some kind of way when we really don't. Still a happy camper here though!

PR #480, opened 2025-08-25
Make squin identity sites an argument rather than an attribute

### LINKED ISSUE #479: Squin identity sites should be an argument instead of an attribute
When you want to "build up" an operator that acts on multiple qubits, you'll often find yourself in a situation where you'll have to write

$$
X_i = \mathbb{1} \otimes \mathbb{1} \otimes ... \otimes X \otimes \mathbb{1} ...
$$

in order to "embed" a single qubit operator on a larger Hilbert space.

In squin, you'd do something like

```python
x = squin.op.x()
id_lhs = squin.op.identity(sites=i)
id_rhs = squin.op.identity(sites=n - i - 1)
x_i = squin.op.kron(id_lhs, squin.op.kron(x, id_rhs))
```

Unfortunately, you can't use this approach unless `n` and `i` are actual constant values. That means you can neither define a short-hand `embed(n: int, idx: int, op: Op)` (which we'd probably want as part of stdlib) nor can you use the above in a for loop. Also, you can't define an operator for different `n`.

To solve this, we'll have to make `sites` an argument rather than an attribute. The obvious downside is that a couple of different analysis will have a more difficult implementation, since you won't always be able to infer the number of sites from the identity anymore.

david-pl:
Closes #479 .

As expected, this makes the implementation of the `NSitesAnalysis` a bit more tricky and also leads to situations where you can't infer the number of sites of an identity.


src/bloqade/squin/analysis/nsites/impls.py
```diff
@@ -88,6 +88,23 @@ def pauli_string(
...
+        sites = stmt.sites
+
+        if not isinstance(sites, ir.ResultValue):
+            return (interp.lattice.top(),)
+
+        if not isinstance(site_stmt := sites.stmt, py.Constant):
+            return (interp.lattice.top(),)
+
+        if not isinstance(value := site_stmt.value, ir.PyAttr):
+            return (interp.lattice.top(),)
```
kaihsin, line 104: this should return bottom instead of top tho

kaihsin: also this does not deal with alias, so double check if the analysis are run before alias inline pass

kaihsin: or, assuming wrap const and cont prop is done before this analysis, then you can get the constant it from hint["const"]

kaihsin (changes requested):

src/bloqade/squin/analysis/nsites/impls.py
david-pl: Yeah, it's not run, which is why there were no hints. But I'll just include it in the run then.

> this should return bottom instead of top tho

Can you explain why? I thought since there is no error, but you just can't know how many sites there are since `sites` is not a constant that top would be the way to go as it's `AnySites`.

kaihsin: just change the get constant as @weinbe58 mentioned below then

weinbe58 (approved):
Just some minor comments

Issue #375, opened 2025-07-15
Lower cirq.TaggedOperation to squin


david-pl:
The `cirq_utils.parallelize` function introduces tags into circuits. Lowering of the resulting circuits to squin fails as we're missing support for `cirq.TaggedOperation`.

PR #489, opened 2025-09-17
Remove cirq tags in parallelize and deal with tags when lowering to squin

### LINKED ISSUE #488: cirq_utils parallelize should remove tags when done
```python
import cirq
from bloqade.cirq_utils import parallelize

q = cirq.LineQubit.range(12)
circuit = cirq.Circuit(
    cirq.X.on_each(q)
)

print(parallelize(circuit))
```

The output of something like this is cluttered with the tags `AUTO:0, AUTO:1, ...`, which is not very nice.

Also related #375: we may support lowering of tagged operations by simply re-using the logic to remove them before the lowering the circuit. I wouldn't be sure what to do with a tag when lowering anyway, so we'd likely just throw it out.

david-pl:
Closes #488 and #375 .


weinbe58 (approved):
LGTM

PR #500, opened 2025-09-19
Remove the custom for loop unroll pass in stim passes

david-pl:
There's still one test that errors and I'm not sure why:

```python
FAILED test/stim/passes/test_squin_qubit_to_stim.py::test_non_pure_loop_iterator - KeyError: <BlockArgument[int] rnd, uses: 1>
````


weinbe58 (approved):
The pass needs to be revisited and probably simplified, but this looks good to me!

Issue #498, opened 2025-09-18
re-implementing the SQUIN -> STIM pass for clifford dialect


Roger-luo:
once `clifford` dialect is implemented (#494) we can start working on this one, I would expect the current pass being simplified a lot. This is orthogonal to the `repeat` statement so let's work on this after `repeat` is done.

Issue #492, opened 2025-09-18
SQUIN refactor tracking issue


Roger-luo:
Let's do this by adding the new dialects first then refactoring existing ones

- [ ] `native` dialect and its stdlib that contains only our native gates
- [ ] `clifford` dialect that contains commonly used clifford gates - `X`, `Y`, `Z`, `S`, `H`, `CZ`, `CNOT` (there are many others we can add later together with extending the native stdlib)
- [ ] refactoring the `noise` dialect to simplify the current definition
- [ ] adding `qubit_id` and `measure_id` into `qubit` dialect
- [ ] re-implementing the SQUIN <-> Cirq pass to this new dialect group
- [ ] re-implementing the SQUIN -> STIM lowering pass

Please help me add other items if they pop up. 

Example of usage

```python
from bloqade.squin import kernel, gate, qubit

@kernel
def s(qubit):
    S(qubit)

@kernel
def s_adj(qubit):
    S(qubit, adjoint=True)

@kernel
def main():
    qubits = qubit.new(3)
    gate.x(qubits[0])
    gate.broadcast.x(qubits)
    
    gate.rz(qubits[0], 0.5) // radians
    gate.broadcast.rz(qubits, 0.5)
```

weinbe58:
We should also think about moving the `squin.qubit` dialect into the top level since it will be shared between squin and native.

PR #510, opened 2025-09-29
Refactor SquinToStimPass for clifford dialect + performance

johnzl-777:
Also included a repetition code - like unit test. I figure this would be nice to have first before working on #498 

I believe this should be backportable too which would be great because I'd like to use the `flatten` pass elsewhere instead of having identical versions floating around (:


johnzl-777:
Considering how breaking everything is for #492 I think it's better to just take care of #498 here as well

PR #511, opened 2025-09-29
Target Lowering `squin.clifford` to `native` dialect group.  

weinbe58:
In this PR I implement a pass and target that rewrite the new `squin.clifford` dialect into the `native` dialect group.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,7 +30,7 @@ jobs:
       - name: Set up Python ${{ matrix.python-version }}
         run: uv python install ${{ matrix.python-version }}
       - name: Install the project
-        run: uv sync --dev --all-extras --no-extra pyqrack-cuda
+        run: uv sync --dev --all-extras --no-extra pyqrack-cuda --no-extra pyqrack-opencl
       - name: Run tests
         # For example, using `pytest`
         run: uv run just coverage
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -17,8 +17,7 @@ dependencies = [
     "rich>=13.9.4",
     "pydantic>=1.3.0,<2.11.0",
     "pandas>=2.2.3",
-    "pyqrack>=1.69.0; sys_platform == 'darwin-arm64'",
-    "pyqrack-cpu>=1.69.0; sys_platform != 'darwin-arm64'",
+    "pyqrack-cpu>=1.69.1",
 ]
 
 [project.optional-dependencies]
@@ -41,10 +40,10 @@ cirq = [
     "qpsolvers[clarabel]>=4.7.0",
 ]
 pyqrack-opencl = [
-    "pyqrack>=1.69.0; sys_platform == 'darwin-arm64'",
+    "pyqrack>=1.69.1",
 ]
 pyqrack-cuda = [
-    "pyqrack-cuda>=1.69.0",
+    "pyqrack-cuda>=1.69.1",
 ]
 stim = [
     "stim>=1.15.0",


src/bloqade/native/upstream/squin2native.py
```diff
@@ -0,0 +1,128 @@
...
+
+        # do not rewrite in invoke because callgraph pass will be looking for invoke statements
+        (callee := py.Constant(native_method)).insert_before(node)
+        node.replace_by(func.Call(callee.result, tuple(node.args), kwargs=()))
+
+        return RewriteResult(has_done_something=True)
+
+
+@dataclass
+class UpdateDialectsOnCallGraph(passes.Pass):
```
david-pl, line 48: This should eventually be upstreamed to kirin. Not needed right now though.

src/bloqade/native/upstream/squin2native.py
```diff
@@ -0,0 +1,128 @@
...
+            return RewriteResult()
+
+        if isinstance(node, stmts.SingleQubitNonHermitianGate):
+            native_method = native_methods[1] if node.adjoint else native_methods[0]
+        else:
+            native_method = native_methods[0]
+
+        # do not rewrite in invoke because callgraph pass will be looking for invoke statements
+        (callee := py.Constant(native_method)).insert_before(node)
+        node.replace_by(func.Call(callee.result, tuple(node.args), kwargs=()))
```
david-pl, line 42: If I understand correctly, you are replacing each statement with an actual `invoke` to a stdlib function. On the one hand, this is quite elegant since it re-uses a lot of code. On the other hand, however, I'm not sure if this leads to the best IR here.

Since we use stdlib everywhere (also in clifford), it's quite likely that the statements you are rewriting are already inside an invoke of a `clifford.stdlib` function. So you will end up nesting a lot of invokes.

Maybe we can inline the newly added invokes here? You could collect the hashes of the invokes you insert and inline them in the end with an appropriate `heuristic`.

src/bloqade/native/upstream/squin2native.py
```diff
@@ -0,0 +1,128 @@
...
+
+    def unsafe_run(self, mt: ir.Method) -> RewriteResult:
+        mt_map = {}
+
+        cg = CallGraph(mt)
+
+        all_methods = set(sum(map(tuple, cg.defs.values()), ()))
+        for original_mt in all_methods:
+            if original_mt is mt:
+                new_mt = original_mt
```
david-pl, line 73: Technically, I'd expect this pass to also update the dialects of `mt`, but I don't think that is happening here (unless I'm missing something). It doesn't matter here, since you are creating a `mt.similar` in `SquinToNative` before calling this pass. Should we upstream it, however, I'd suggest to change this.

src/bloqade/native/upstream/squin2native.py
```diff
@@ -0,0 +1,128 @@
...
+        new_dialects = mt.dialects.discard(clifford_dialect).union(kernel)
+
+        out = mt.similar(new_dialects)
+        UpdateDialectsOnCallGraph(new_dialects, no_raise=no_raise)(out)
+        SquinToNativePass(new_dialects, no_raise=no_raise)(out)
+        # verify all kernels in the callgraph
+        new_callgraph = CallGraph(out)
+        all_kernels = (ker for kers in new_callgraph.defs.values() for ker in kers)
+        for ker in all_kernels:
+            ker.verify()
```
david-pl, line 126: Wait, wouldn't `out.verify()` do this? At least, I'd say that _should_ be the behavior. Maybe a kirin issue?

test/native/upstream/test_squin2native.py
```diff
@@ -0,0 +1,42 @@
...
+    def main():
+        q = squin.qubit.new(n)
+        squin.h(q[0])
+
+        for i in range(n - 1):
+            squin.cx(q[i], q[i + 1])
+
+        squin.broadcast.sqrt_x_adj(q)
+
+    new_main = SquinToNative().emit(main)
```
david-pl, line 25:
```suggestion
    new_main = SquinToNative().emit(main, no_raise=False)
```

Just to be sure, I'd always set that in tests.

david-pl (approved):
Looks good! Just some minor comments / suggestions.

src/bloqade/native/upstream/squin2native.py
weinbe58: Actually this is more efficient if you are only using the clifford standard libraries I will rewrite those subroutines and not the invoke of the subroutine. the complexity is bounded by the number of standard library functions. not by the number of invokes of those standard library functions.

src/bloqade/native/upstream/squin2native.py
weinbe58: yeah as discussed on slack ,Passes to not copy the method but they can (and should in this case) copy data inside the method

weinbe58: There is a copy inside the `SquinToNative.emit` method.

Issue #341, opened 2025-06-30
Syntax sugar for squin's two qubit pauli channel


david-pl:
Currenlty, the wrapper takes a list of exactly 15 parameters, e.g.

```python
@squin.kernel
def main():
    q = squin.qubit.new(2)
    two_qubit_noise = squin.noise.two_qubit_pauli_channel(
        [
            0.036,
            0.007,
            0.035,
            0.022,
            0.063,
            0.024,
            0.006,
            0.033,
            0.014,
            0.019,
            0.023,
            0.058,
            0.0,
            0.0,
            0.064,
        ]
    )
    squin.qubit.apply(two_qubit_noise, q)
```

This is not very user friendly, especially when you have a lot of `0` values in there. Also, it's easy to make mistakes.

I'd suggest we add syntax sugar that allows passing a `dict` similar to cirq's `error_probabilities` in the `AsymmetricDepolarizingChannel`, so you could just write e.g.

```python
@squin.kernel
def main():
    q = squin.qubit.new(2)
    two_qubit_noise = squin.noise.two_qubit_pauli_channel({'XX': 0.1, 'YY': 0.2, 'ZZ': 0.15})
    squin.qubit.apply(two_qubit_noise, q)
```

Anything omitted here should default to 0.

PR #512, opened 2025-09-30
Refactor & simplify noise dialect

david-pl:
This simplifies the noise dialect a lot. I also cut down the supported semantics, so we now only support one- and two-qubit noise channels. Effectively, we now only have the following statements:

* `SingleQubitPauliChannel`
* `TwoQubitPauliChannel`
* `Depolarize`
* `Depolarize2`
* `QubitLoss`

Technically, `Depolarize(2)` and `Single(Two)QubitPauliChannel` are redundant in that you can always rewrite `Depolarize` to the other channel. But I didn't want to burden the squin to stim pass with matching that (you'd need to look at runtime values of probabilities to make sure things are symmetric) when rewriting to `Depolarize`. At the same time, we need the asymmetric versions for cirq lowering.

One thing I realized: going the same stdlib approach as for the native and the clifford dialects, we end up needing `broadcast` again. And I didn't want to have `squin.channel.broadcast.depolarize` and similar things. So, for now, I just imported everything to the top level again, which means we have `squin.depolarize` and `squin.broadcast.depolarize`.

FYI @johnzl-777 I renamed the STIM tests, so they'll be skipped in the CI. If this gets merged before you get to the re-implementation of the squin to stim pass, keep that in mind.

Finally, I want to point out that it's not really possible to make the linter happy when going through a stdlib function that includes a list of probabilities, specifically `two_qubit_pauli_channel`. The problem is, that the linter complains that `[0.01, 0.02, ...]` isn't an `ilist`, but it's really lowered to one (the linter just doesn't know), so changing the type hint in the stdlib function to `list[float]` errors since it's a kernel. That said, we need some special handling anyway to improve the API here (see also #341 ). But let's do this in another PR.


weinbe58 (approved):
LGTM

PR #518, opened 2025-09-30
Simplyfing `qubit.new`

### LINKED ISSUE #508: simplifying `qubit.new`
- [x] add a new statement `qubit.NewQubit` now only allocates a **single** qubit
- [ ] create a stdlib function `qubit.new(int)` that allocates `IList[Qubit, N]`
- [x] create a wrapper `qubit.new_qubit()` that creates a single new qubit.

Caveat:

We don't have dependent types so when `%n` is non-constant at compile-time in `qubit.new(%n)` we won't be able to infer a variable length as type variable. This is prob ok for now because we don't have dynamic allocated qubits (or just a few ancilla that does not require a list)

weinbe58:
Closes #508 however I am not sure where to reexport `new` since I Can't reexport it into the `qubit` module. For now I have re-exported it into `squin.new`.

### DIFF
--- a/src/bloqade/squin/qubit.py
+++ b/src/bloqade/squin/qubit.py
@@ -30,6 +30,12 @@ class New(ir.Statement):
 
 
 @statement(dialect=dialect)
+class NewQubit(ir.Statement):
+    traits = frozenset({lowering.FromPythonCall()})
+    result: ir.ResultValue = info.result(QubitType)
+
+
+@statement(dialect=dialect)
 class Apply(ir.Statement):
     traits = frozenset({lowering.FromPythonCall()})
     operator: ir.SSAValue = info.argument(OpType)
@@ -102,6 +108,16 @@ def new(n_qubits: int) -> ilist.IList[Qubit, Any]:
 
     Returns:
         (ilist.IList[Qubit, n_qubits]) A list of qubits.
+    """
+    ...
+
+
+@wraps(NewQubit)
+def new_qubit() -> Qubit:
+    """Create a new qubit.
+
+    Returns:
+        Qubit: A new qubit.
     """
     ...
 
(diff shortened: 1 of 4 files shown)


src/bloqade/squin/qubit.py
```diff
@@ -106,6 +112,16 @@ def new(n_qubits: int) -> ilist.IList[Qubit, Any]:
     ...
 
 
+@wraps(NewQubit)
+def new_qubit() -> Qubit:
```
Roger-luo, line 116: rename `NewQubit` -> `New`

src/bloqade/cirq_utils/emit/base.py
```diff
@@ -112,10 +114,44 @@ def main():
...
+
+    callable_region = mt.callable_region.clone()
+    entry_block = callable_region.blocks[0]
+    args_ssa = list(entry_block.args)
+    first_stmt = entry_block.first_stmt
+
+    assert first_stmt is not None, "Method has no statements!"
+    if len(args_ssa) - 1 != len(args):
+        raise EmitError(
+            f"The method {sym_name} takes {len(args_ssa)} arguments, but you passed in {len(args)} via the `args` keyword!"
```
david-pl: Since you're comparing `len(args_ssa) - 1 == len(args)` here, this is gonna be another one of those nice errors complaining that `2 arguments are expected, but 2 were given`, right?

```suggestion
            f"The method {sym_name} takes {len(args_ssa) - 1} arguments, but you passed in {len(args)} via the `args` keyword!"
```

src/bloqade/squin/groups.py
```diff
@@ -20,13 +20,14 @@ def run_pass(method: ir.Method, *, fold=True, typeinfer=True):
...
-            typeinfer_pass(method)
+            typeinfer_pass(method)  # infer types before desugaring
             desugar_pass.rewrite(method.code)
 
         ilist_desugar_pass(method)
 
         if typeinfer:
             typeinfer_pass(method)  # fix types after desugaring
             method.verify_type()
+            # method.print()
```
weinbe58:
```suggestion
```

Issue #495, opened 2025-09-18
refactoring the noise dialect to simplify the current definition


Roger-luo:
this will be breaking, we should touch it after we finish the #494 and #493

PR #520, opened 2025-10-01
Refactor cirq - squin emit and lowering

### LINKED ISSUE #497: re-implementing the SQUIN <-> Cirq pass
re-implementing the SQUIN <-> Cirq pass once all previous dialects are finished. Potentially with some thinking about how to turn part of this into a tutorial so we can show people how integration could work.

david-pl:
Closes #497 

I had to cut down support for some cirq features in lowering, but I don't think they were used heavily. For example, we no longer have `CZPowGate` with arbitrary exponents or arbitrary controlled operations.

Also, one more significant breaking change is that `func.Invoke` statements are no longer emitted as `CircuitOperation`, but are added to the circuit flat-out. I think this matches user expectations and otherwise would have been really odd given our new approach with stdlib functions.

Finally, noise still needs to be done, but that is pending #495, which should be merged first, I think.

**Edit**: Noise is also done now, but I also had to cut support for `AmplitudeDampingChannel`, since we can't really represent this with the current statements in the noise dialect (I think). However, I haven't seen this used anywhere so far, so I'd suggest we wait for someone to actually require that feature.


weinbe58 (approved):
LGTM, @Roger-luo, I know we had discussed doc-strings on the new API, but here all the visitor methods and interpreter methods are not Private. Should we mark the classes as private in this case?

david-pl:
@weinbe58 @Roger-luo I went ahead and marked the method tables as private, since there's really no point in having those public. I'm not so sure about the target classes themselves though, so let me know how to proceed here.

I'll still go ahead and merge so as to not block other progress in the refactor. We can still change this during the refactor to make it private or add docstrings.

PR #525, opened 2025-10-07
Fix for loop unrolling in qasm2 emit

david-pl:
This bug was introduced with the new `UnrollScf` pass in kirin v0.17.23. It currently breaks the QASM2 GHZ example in the documentation.

@weinbe58 @Roger-luo I'm not entirely sure we still need the `trim.UnusedYield` here, all tests and examples work without it. Still, I left it in just to be sure.

### DIFF
--- a/src/bloqade/qasm2/passes/fold.py
+++ b/src/bloqade/qasm2/passes/fold.py
@@ -20,6 +20,7 @@ from kirin.analysis import const
 from kirin.dialects import scf, ilist
 from kirin.ir.method import Method
 from kirin.rewrite.abc import RewriteResult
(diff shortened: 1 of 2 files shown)


david-pl:
> The one use case for having the Trim rewrite is to pair it with a LiftIf rewrite because you can safely lift any ssa result that is not within the yield of the IfElse bodies.

@weinbe58 good to know, thanks.

> @david-pl do you want me to look at the GHZ example?

What do you mean? I added the part that was not working as a unit test, is that not sufficient?

david-pl:
Superseded by #536

PR #526, opened 2025-10-07
correlated loss support

rafaelha:
This pull request adds support for correlated qubit loss noise channel to
- the `squin.noise` dialect and associated stdlib
- `stim` dialect
- PyQrack interpreter


src/bloqade/stim/dialects/noise/stmts.py
```diff
@@ -89,8 +89,8 @@ class NonStimError(ir.Statement):
 class NonStimCorrelatedError(ir.Statement):
     name = "NonStimCorrelatedError"
     traits = frozenset({lowering.FromPythonCall()})
-    nonce: int = (
-        info.attribute()
+    nonce: int = info.attribute(
```
rafaelha, line 92: I think the correlated error previously used to exist in the code base. I'm wondering if the previous approach also assigned a random number to `nonce` or if there is a better solution.

rafaelha:
Coverage is failing but that's due to the skipped files (due to ongoing refactor).

A test in `bloqade-circuit/test/stim/passes/squin_noise_to_stim.py` still needs to be added, but it might be best to wait until that file is refactored.

src/bloqade/squin/stdlib/simple/noise.py
```diff
@@ -97,6 +97,20 @@ def qubit_loss(p: float, qubit: Qubit) -> None:
...
+    """
+    Apply a correlated qubit loss channel to the given qubits.
+
+    All qubits are lost together with a probability `p`.
+
+    Args:
+        p (float): Probability of the qubits being lost.
+        qubits (IList[Qubit, Any]): The list of qubits to which the correlated noise channel is applied.
+    """
+    broadcast.correlated_qubit_loss(p, qubits)
```
rafaelha, line 111: This is somewhat hacky. `correlated_qubit_loss` is an `n`-qubit operator. As far as I know all other operators act on either one or two qubits.

Stim does not support `n`-qubit operators, which is why the `nonce` workaround is introduced.

I'm wondering if `broadcast` should be disallowed for this type of operator?  Instead only use `apply`? I am a bit confused how that works after the refactor.

src/bloqade/stim/dialects/noise/stmts.py
cduck: The nonce solution was mine originally.  If you can find a stim instruction other than I_ERROR that satisfies the following, then please use it.
- The instruction as emitted from kirin must have no effect when simulated in stim
- The instruction must maintain groups of qubits, either within a single instruction or by the stim library not merging multiple instructions in sequence.

src/bloqade/squin/stdlib/simple/noise.py
cduck: This mismatch in semantics is a concern.  There are a couple stim gates that act on more qubits: MPP, [SPP](https://github.com/quantumlib/Stim/blob/main/doc/gates.md#SPP), and SPP_DAG
```
# Perform a SQRT_YY gate between qubit 1 and 2, and a SQRT_ZZ_DAG between qubit 3 and 4.
SPP Y1*Y2 !Z1*Z2
```

src/bloqade/stim/dialects/noise/stmts.py
rafaelha: I went with `I_ERROR[correlated_loss:<nonce>]` after talking to @kaihsin. In the `SquinToStimPass`, the nonce value is just incremented for every correleated_loss operation.

If you use the stim dialect directly, you can either specify the nonce value or it will be assigned a random number. I find that defaulting to a random value helps prevent unintended side effects — but I’m open to changing the default.

src/bloqade/squin/noise/stmts.py
```diff
@@ -24,6 +24,11 @@ class TwoQubitNoiseChannel(NoiseChannel):
     pass
 
 
+@statement
+class MultiQubitNoiseChannel(NoiseChannel):
```
david-pl: You don't need this base class since you're adding just a single statement anyway.

src/bloqade/stim/rewrite/squin_noise.py
```diff
@@ -31,6 +34,28 @@ def rewrite_NoiseChannel(
         if rewrite_method is None:
             return RewriteResult()
 
+        if isinstance(stmt, squin_noise.stmts.MultiQubitNoiseChannel):
```
david-pl: If you remove the base class, remember to update this check here as well.

src/bloqade/stim/rewrite/squin_noise.py
```diff
@@ -31,6 +34,28 @@ def rewrite_NoiseChannel(
...
+            if not isinstance(qubit_address_attr, AddressAttribute):
+                return RewriteResult()
+
+            address_tuple = qubit_address_attr.address
+
+            if not isinstance(address_tuple, AddressTuple):
+                return RewriteResult()
+
+            for address in address_tuple.data:
+                qubit_idx_ssas = insert_qubit_idx_from_address(
```
david-pl: Can `qubit_idx_ssas` be `None`? If so, you need to check for that and bail if it is `None`.

src/bloqade/squin/noise/stmts.py
rafaelha: Ok, I removed it. We can always add it back if we add more channels in the future.

src/bloqade/stim/dialects/noise/stmts.py
```diff
@@ -89,8 +89,8 @@ class NonStimError(ir.Statement):
...
+        default_factory=lambda: __import__("random").getrandbits(32)
```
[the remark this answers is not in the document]
johnzl-777: Is there a reason we don't want it to be 0 if we write in the stim dialect? 

I should mention the stim dialect was never intended to be a user-facing dialect, it's primarily a target for converting from other dialects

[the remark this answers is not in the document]
cduck: Um, default nonce=0 is the worst idea because it will collide by default!  I would at least make it a required parameter so whoever uses the instruction is forced to find a way to determine non-colliding nonce values.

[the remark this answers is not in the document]
cduck: Unless stim-emit replaces 0-nonces with a counter during the emit step?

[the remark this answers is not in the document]
johnzl-777: Oh wait yeah I'm mixing up the logic of the rewrite, if you're spawning individual stim dialect statements from scratch then yes, one should avoid a hardcoded nonce. My bad!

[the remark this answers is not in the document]
kaihsin: @cduck I think 

1. The default behavior for stim dialect should be always collide (a default value) 
2. For from squin -> stim rewrite, a global counter will properly lower to stim statement with non-colliding index by default
3. For user directly writing low level stim dialect, its on user's responsibility to assign unique integer to avoid this

Since stim is a target dialect, 1 and 3 behavior align with typical way user write stim text format too

PR #534, opened 2025-10-08
Add U3 gate statement to squin.gate

### LINKED ISSUE #532: U3 gate in `squin.gate` dialect.
In order to support the Gemini product offerings, we must include a `U3` gate in the IR as a statement to make sure we can represent all valid programs on the statement level.

david-pl:
Closes #532 .

@weinbe58 @johnzl-777 Before merging, we need to clarify the convention we are using for U3.

Both PyQrack and QASM2 use the definition

$$ U3(\theta, \phi, \lambda) = R_z(\phi)R_y(\theta)R_z(\lambda) = R_y(\phi + \frac{\pi}{2})R_x(\theta)R_z(\lambda - \frac{\pi}{2}) $$

whereas qiskit and the stdlib we had before differ by a phase factor

$$ e^{(\phi + \lambda) / 2} R_z(\phi)R_y(\theta)R_z(\lambda) $$

Which one should we use? For now, I stuck with the first one, mainly because I can just forward to the PyQrack `sim_reg.u` method without modifications. We either need to change that, or the standard library definitions. Since it's a statement, this should also be clearly communicated, so I'd update the docstrings too, once we made a decision.

### DIFF
--- a/src/bloqade/cirq_utils/lowering.py
+++ b/src/bloqade/cirq_utils/lowering.py
@@ -169,7 +169,6 @@ DecomposeNode = (
     | cirq.CSwapGate
     | cirq.XXPowGate
     | cirq.YYPowGate
-    | cirq.ZZPowGate
     | cirq.CCXPowGate
     | cirq.CCZPowGate
 )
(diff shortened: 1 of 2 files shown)


kaihsin:
I don't have strong preference but just reminder that the u3 to Clifford rewrite, and downstreams that use squin would all be breaking, and has to change accordingly.

weinbe58:
Good question, In the native gate library I have the extra phase added so let's use that convention

weinbe58:
@kaihsin The second one, the one with the extra phase.

kaihsin:
then no. I think majority of the convention in our pipeline using the first one

PR #536, opened 2025-10-08
Refactor `QASM2Fold` and `CanonicalizeIList`

weinbe58:
Just doing some small housekeeping using the new passes/rewrite rules added in Kirin in the past few months.


david-pl:
@weinbe58 no worries, that's not what I meant 😄 

This PR is cleaner than my fix, so by all means let's use this one. FYI, I tried and it also fixes the issue I had in #525 .
Still, it would make sense to add this as unit test, I think, since it's still failing on main.

Issue #545, opened 2025-10-14
Reset statement wrapper


johnzl-777:
#510 Brings back the `Reset` statement (as well as dropping `ResetToOne`) but this statement does not have a wrapper that conforms to the latest operator semantics. 

I had tried in that PR to just refactor `qubit` to follow the format of the `gate` dialect but that broke enough old tests I felt it warranted a separate PR instead of being shoehorned into #510. 

Note: there is a test in #547 that has a local wrapper for reset that should be removed once this is properly implemented (:

PR #547, opened 2025-10-15
Implement cirq and PyQrack methods for reset stmt

david-pl:
For testing purposes I added a wrapper in the unit test since we currently don't have one. This is to be replaced once we have #545 .


johnzl-777 (approved):
Looks good, thanks for doing this! 

Had a hell of time trying to figure out the PyQrack API on reset, I had some vague recollection of that measurement + X method from before

Issue #549, opened 2025-10-16
RFC: qubit dialect structure conformity with gate/broadcasting


johnzl-777:
Currently the `qubit` dialect, under the `squin` umbrella, contains key statements for writing squin programs but makes UX a bit odd (unless you explicitly import `qubit` as a separate thing but this harkens back to https://github.com/QuEraComputing/bloqade/issues/287) 

```python
from bloqade import squin

@squin.kernel
def prog():
	q = squin.qubit.new(10)
	squin.x(q[1])
	squin.broadcast.h(q)
	measurement_results = squin.qubit.measure(q)
```

It would be much nicer if a user could just do `squin.measure` instead. I understand post- #518 the `qubit.new` will be a nicer `squin.qalloc`.

I realize a particularly easy way would be to just do some kind of re-export but then if we have operations like `Reset` (pending a wrapper: #545 - one that I'd like to add post conclusion on this RFC) is it best they follow the split we have between single and broadcasted qubit operations? Or should `Reset` follow something like `Measure` with a `MeasureAny` blanket statement + desugaring?

I think of a similar issue for `get_qubit_id` and `get_measurement_id`. Right now they only accept singular instances of Qubits and MeasurementResults. Do we want them to be broadcastable too? Like

```python
qs = squin.qubit.new(10) 

ids : IList[int] = get_qubit_id(qs)
# but could also do 
id: int = get_qubit_id(q[0])
```

weinbe58:
sounds good. also reexporting will be easier once qubit is at the top level anyways.

johnzl-777:
Yeah that sounds pretty reasonable to me! I suppose my final question is the following:

for something like `get_qubit_id(qubit) -> int` should _that_ be broadcastable too considering it consumes qubits? And if we make that broadcastable should `get_measure_id(MeasurementResult) -> int` have something similar?

david-pl:
As a side note: if we change the semantics to `squin.measure(q: Qubit)` and `squin.broadcast.measure(q: IList[Qubit])`, we can get rid of the `MeasureQubit` statement (basically replace it by `MeasureQubitList`) and the `MeasureAny` statements along with desugaring.
From past experiences with `ApplyAny` and friends, I'd say this is another argument as to why this would be beneficial to have since it simplifies the dialect and avoids situations where the desugaring fails leading to implementation errors down the line.

[...]

PR #557, opened 2025-10-21
Restructure and move qubit dialect

david-pl:
As discussed in #549, this restructures the qubit dialect. I also moved it to the top level of bloqade, since it's used in multiple places. It's still re-exported by squin, though.

I've got it down to a single failing test, where the `AggressiveUnroll` fails because the constant propagation doesn't hint the loop iterable as constant for some reason.

This breaks the syntax such that you have to change the function calls using lists such as `squin.qubit.measure(q: IList[Qubit])` to `squin.broadcast.measure(q: IList[Qubit])`.

**Edit**: turns out I was just too eager to remove the type inference method @weinbe58 added in #508. That resolves the failing test.

**Edit2**: I've also changed `QubitId` and `MeasurementId` to be "broadcasted" (i.e., they now operate on lists of qubits / measurements and return lists). This means that now **all statements except `qubit.new` are broadcasted**, which is as consistent as we can have it, I think.

### DIFF
--- a/src/bloqade/squin/groups.py
+++ b/src/bloqade/squin/groups.py
@@ -1,12 +1,12 @@
 from kirin import ir, passes
 from kirin.prelude import structural_no_opt
-from kirin.dialects import ilist
+from kirin.dialects import debug, ilist
 
 from . import gate, noise
 from .. import qubit
 
 
-@ir.dialect_group(structural_no_opt.union([qubit, noise, gate]))
+@ir.dialect_group(structural_no_opt.union([qubit, noise, gate, debug]))
 def kernel(self):
     fold_pass = passes.Fold(self)
     typeinfer_pass = passes.TypeInfer(self)
--- a/src/bloqade/stim/groups.py
+++ b/src/bloqade/stim/groups.py
@@ -1,12 +1,12 @@
 from kirin import ir
 from kirin.passes import Fold, TypeInfer
-from kirin.dialects import func, lowering
+from kirin.dialects import func, debug, lowering
 
 from .dialects import gate, noise, collapse, auxiliary
 
 
 @ir.dialect_group(
-    [noise, gate, auxiliary, collapse, func, lowering.func, lowering.call]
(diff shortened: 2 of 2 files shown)


src/bloqade/qubit/stdlib/broadcast.py
```diff
@@ -0,0 +1,62 @@
...
+@kernel
+def measure(qubits: ilist.IList[Qubit, N]) -> ilist.IList[MeasurementResult, N]:
+    """Measure a list of qubits.
+
+    Args:
+        qubits (IList[Qubit, N]): The list of qubits to measure.
+
+    Returns:
+        IList[MeasurementResult, N]: The list containing the results of the measurements.
+            A MeasurementResult can represent both 0 and 1, but also atoms that are lost.
```
johnzl-777:
```suggestion
            A MeasurementResult can represent both 0 and 1 as well as atom loss.
```
This is a pretty minor nitpick, I know you have this in multiple parts of the docstrings. I think this just flows a bit better in terms of readability.

src/bloqade/qubit/_prelude.py
```diff
@@ -0,0 +1,43 @@
...
+            bool,
+            Doc(
+                "run type inference and apply the inferred type to IR, default `False`"
+            ),
+        ] = False,
+        fold: Annotated[bool, Doc("run folding passes")] = True,
+        aggressive: Annotated[
+            bool, Doc("run aggressive folding passes if `fold=True`")
+        ] = False,
+        no_raise: Annotated[bool, Doc("do not raise exception during analysis")] = True,
```
johnzl-777: I assume for the previous annotations if you mention the default value, you should carry that pattern over here too?

johnzl-777 (approved):
LGTM! Just a dangling print statement and some (very) minor nitpicks. Thank you for doing this!

david-pl:
CI is now failing because of #566 :/

Issue #531, opened 2025-10-07
Gemini logical target dialect group in `bloqade.gemini`


weinbe58:
We should restrict which dialects we allow inside hardware to a smaller dialect group for verification purposes before lowering continues.

The dialects to include are:

* `py.constant`
* `qubit`
* `squin.gates`
* `ilist`
* `func`

PR #558, opened 2025-10-21
Implement gemini.logical and draft kernel validation

david-pl:
This implements the simple dialect group as discussed in #531.

However, it also contains a draft (RFC) for validating kernels, using the above as an example.

I thought a bit about it and in the end decided to split the validation into two parts:

1. A `ValidationAnalysis` pass with an appropriate lattice (could be extended to warnings, etc.) that checks for errors in the IR.
2. A simple `KernelValidation` class, that gathers up the errors and throws them as `kirin.ir.ValidationError` with the appropriate source info pointing to the line of the error. I went back and forth as to how to implement this, since it is a bit similar to interpretation. However, in the end I just decided to go with a very simple dataclass, since it is sufficiently different from an interpreter or a rewrite.

One thing that's not great and I'm not sure how to improve: we probably don't want the `ValidationAnalysis` to implement a lot of methods via method tables. For example, the `gemini.logical` dialect doesn't support `scf.IfElse`. However, if we want to re-use the `ValidationAnalysis` elsewhere, then registering a `MethodTable` for `scf` with a method for `scf.IfElse` might fail validation in kernels where this should be fine. At the same time, we might want to share validation between different kernels.

The way I "solved" this for now is to define a new analysis pass just for the logical dialect that inherits from the generic `ValidationAnalysis`. If we go this route, then the `ValidationAnalysis` basically just defines the lattice and a fallback method.

There's also some TODOs left here, e.g. how to display multiple errors, but I thought I'd get comments as soon as possible.


src/bloqade/gemini/analysis/logical_validation/impls.py
```diff
@@ -0,0 +1,64 @@
...
+        stmt: qubit.New,
+    ):
+        # TODO: this is actually tricky, since qalloc calls qubit.new multiple times and we have to make sure qalloc is only called once
+        # but it can technically contain many qubit.new calls
+        # if interp.has_allocated_qubits:
+        #     raise ir.ValidationError(
+        #         stmt, "Can only allocate qubits once in a logical Gemini program!"
+        #     )
+
+        # interp.has_allocated_qubits = True
```
weinbe58: Actually what matters is if qubit are allocated after the first gate so maybe add a flag for if there ways already a gate operation

weinbe58: Actually I'm not even sure if its a problem if there are allocations after the first gate.

PR #560, opened 2025-10-21
Remove correlated error nonce from stim IR

rafaelha:
This PR removes the `nonce` arg from correlated loss statements. `I_Error` gates in stim need to be tagged with a unique value to prevent merging of gate statements by stim.

Previously, these tags were mirrored in the stim IR, even though our IR does not need them (since there is not auto-merging of statements like in stim).

Now nonce-tags have been removed from the IR. Instead `EmitStimMain` adds them during codegen.


kaihsin (changes requested):
LGTM just some minor comments.

PR #563, opened 2025-10-22
New Address Analysis. 

weinbe58:
~In this PR I implement a Joint analysis with constant prop and Address analysis to try to get the address analysis to work properly going into the call stack.~

I do not use a Cartesian product for the lattic because I need to implement partial lambda and partial IList and tuple to get the analysis to work. In order to support constant folding, however, I opted into adding an extra lattice element to wrap the constant prop results.

### DIFF
--- a/src/bloqade/native/_prelude.py
+++ b/src/bloqade/native/_prelude.py
@@ -7,10 +7,10 @@ from typing_extensions import Doc
 
 from bloqade import qubit
 
-from .dialects import gates
+from .dialects import gate
 
 
-@ir.dialect_group(structural_no_opt.union([gates, qubit]))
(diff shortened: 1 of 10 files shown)


src/bloqade/cirq_utils/lowering.py
```diff
@@ -151,6 +151,10 @@ def main():
         dialects=dialects,
         code=code,
     )
+    assert (run_pass := kernel.run_pass) is not None
+    run_pass(mt, typeinfer=True)
+
+    return mt
 
     assert (run_pass := kernel.run_pass) is not None
```
david-pl: Unreachable code here.

PR #590, opened 2025-10-28
Kirin upgrade: fix naming and argument conventions in `native.gate` dialect. 

weinbe58:
As we're introducing a large breaking change it is probably a good time to update the statements in `native` to be more consistent with `squin`'s naming/call conventions. 

blocked by CI at the moment.

### DIFF
--- a/src/bloqade/native/dialects/gate/__init__.py
+++ b/src/bloqade/native/dialects/gate/__init__.py
@@ -1,2 +1,2 @@
+from . import stmts as stmts
 from ._dialect import dialect as dialect
-from ._interface import r as r, cz as cz, rz as rz


david-pl:
@weinbe58 eh, CI is not happy with the branch. Honestly it may be easier to just start off of main and redo the changes. Or we can leave it for v0.9 together with the kirin upgrade if it's too much of a hassle.

src/bloqade/pyqrack/native.py
```diff
@@ -7,42 +7,42 @@
 from pyqrack import Pauli
 from bloqade.pyqrack import PyQrackQubit
 from bloqade.pyqrack.base import PyQrackInterpreter
-from bloqade.native.dialects import gates
+from bloqade.native.dialects import gate
 
 
-@gates.dialect.register(key="pyqrack")
+@gate.dialect.register(key="pyqrack")
 class NativeMethods(interp.MethodTable):
```
david-pl: I suppose this class should technically be marked as private.

david-pl (approved):
Looks good! Just one very minor comment, but you can ignore that if you want.

One general question: should we explicitly export statements for dialects? Right now, you have e.g. `native.gate.CZ`. Elsewhere, I think we usually only export `stmts`, so statements can be accessed via `native.gate.stmts.CZ`. I think this makes it clearer to users not to access that when trying to call a `CZ`. Then, again, since we now do everything via stdlibs it might not be needed.

PR #602, opened 2025-11-05
Revert "Backport cleaner squin gate generation"

david-pl:
Reverts QuEraComputing/bloqade-circuit#598


johnzl-777:
@cduck I think there was an agreement after a discussion with @david-pl and @kaihsin  this is technically breaking (Kai would have to update things on his end) and this should be avoided unless it's critical 0.7 has the same gate generation features

PR #593, opened 2025-10-30
Adjust AggressiveUnroll to fix type error during unroll

cduck:
I'm not sure what was causing the error but rearranging the passes with the help of @weinbe58 fixed the problem.

For context, this was the error.  I did not find a MWE.
```python
Traceback (most recent call last):
  File "my_script.py", line 49, in compile_to_stim
    PhysicalAndSquinToStim(main.dialects, no_raise=no_raise)(main)
  File "site-packages/kirin/passes/abc.py", line 30, in __call__
    result = self.unsafe_run(mt)
  File "squin_and_to_stim.py", line 99, in unsafe_run
    .unsafe_run(mt)
  File "site-packages/bloqade/rewrite/passes/canonicalize_ilist.py", line 29, in unsafe_run
    ).rewrite(mt.code)
  File "site-packages/kirin/rewrite/fixpoint.py", line 24, in rewrite
    result = self.rule.rewrite(node)
  File "site-packages/kirin/rewrite/walk.py", line 40, in rewrite
    result = self.rule.rewrite(subnode)
  File "site-packages/kirin/rewrite/chain.py", line 29, in rewrite
    result = rule.rewrite(node)
  File "site-packages/kirin/rewrite/abc.py", line 38, in rewrite
    return self.rewrite_Statement(cast(Statement, node))
  File "site-packages/kirin/dialects/ilist/rewrite/flatten_add.py", line 49, in rewrite_Statement
    assert isinstance(rhs_type := rhs.type, types.Generic), "Impossible"
AssertionError: Impossible
```

PR #621, opened 2025-11-11
Fix AggressiveUnroll again

cduck:
Do not merge until we understand the issue being fixed here.  This became an issue again while upgrading my code for bloqade-circuit v0.9.

This pass has been unreliable before (see #593) and I'm not sure why adding a second copy of the `ilist.rewrite.Unroll` pass fixes the issue.

Here's the error it is fixing:
```python
  File "test.py", line 59, in compile_to_stim
    SquinToStimPass(main.dialects, no_raise=no_raise)(main)
  File "kirin/passes/abc.py", line 30, in __call__
    result = self.unsafe_run(mt)
  File "bloqade/stim/passes/squin_to_stim.py", line 71, in unsafe_run
    .rewrite(mt.code)
  File "kirin/rewrite/chain.py", line 29, in rewrite
    result = rule.rewrite(node)
  File "kirin/rewrite/walk.py", line 40, in rewrite
    result = self.rule.rewrite(subnode)
  File "kirin/rewrite/abc.py", line 38, in rewrite
    return self.rewrite_Statement(cast(Statement, node))
  File "bloqade/stim/rewrite/set_observable_to_stim.py", line 26, in rewrite_Statement
    return self.rewrite_SetObservable(node)
  File "bloqade/stim/rewrite/set_observable_to_stim.py", line 39, in rewrite_SetObservable
    measure_ids = self.measure_id_frame.entries[node.measurements]
KeyError: <ResultValue[IList[MeasurementResult, AnyType()]] inputs, uses: 1>
```

Tagging: @kaihsin

### DIFF
--- a/src/bloqade/rewrite/passes/aggressive_unroll.py
+++ b/src/bloqade/rewrite/passes/aggressive_unroll.py
@@ -71,7 +71,9 @@ class AggressiveUnroll(Pass):
     def unsafe_run(self, mt: Method) -> RewriteResult:
         result = RewriteResult()
         result = self.fold.unsafe_run(mt).join(result)
-        result = Walk(ilist.rewrite.Unroll()).rewrite(mt.code).join(result)  # BUG: cduck's use case fails without this line
+        result = (
+            Walk(ilist.rewrite.Unroll()).rewrite(mt.code).join(result)
+        )  # BUG: cduck's use case fails without this line
         result = self.scf_unroll.unsafe_run(mt).join(result)
         self.typeinfer.unsafe_run(
             mt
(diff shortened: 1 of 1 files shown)


kaihsin:
Is there an issue to track this? I can dig a bit more on this since it's affecting downstream

cduck:
> The problem here seems to be that the AggressiveUnroll fails to inline some lambda functions.

This makes sense.  Let's open an issue to fix AggressiveUnroll.  I am confused how the line I added here changes the behavior, because it is a duplicate copy of the ilist.rewrite.Unroll pass but run earlier in the sequence.

kaihsin:
For some reason typeinfer marked this as bottom:
```
%0 = py.constant.constant IList(range(0, 4)) : !Bottom
```

weinbe58:
My bad, accidentally closed the PR

Issue #609, opened 2025-11-07
Create `TerminalMeasure` statement


weinbe58:
see parent issue for details. This is tracking the creation of the statement. This statement needs to live in the gemini logical dialect group.

PR #627, opened 2025-11-18
TerminalLogicalMeasurement implementation

johnzl-777:
I believe the signature should be in line with what was requested in #609 but I went ahead and made a dedicated dialect for the additional statement. 

I originally wanted to follow something like `native`'s structure which has its own version of the squin `qubit` dialect (same name but tucked away elsewhere) although I wasn't sure if this was something we wanted to mirror.

### DIFF
--- a/src/bloqade/gemini/dialects/logical/_interface.py
+++ b/src/bloqade/gemini/dialects/logical/_interface.py
@@ -1,4 +1,4 @@
-from typing import Any, TypeVar
+from typing import TypeVar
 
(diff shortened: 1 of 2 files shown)


johnzl-777:
@david-pl @weinbe58 sorry to re-requst a review here, I kept running into this obnoxious circular import problem but I finally fixed it. The problem though is that "groups" now lives inside logical to keep the `gemini.logical.kernel` structure which kind of irks me.

PR #630, opened 2025-11-20
Stim REPEAT support 

johnzl-777:
This PR enables scf.For statements with the proper format to be represented as a Stim REPEAT statement. There can be no dependence on the iterator variable and the number of iterations must be known at the time of the rewrite.

Per feedback I received from the team presentation and some additional help from Phillip I’ve been able to eliminate the dependency on CSE ordering. CSE can now be invoked at any time in the pass with no impact on the correctness of the resulting program.

That being said, there are definitely areas I’d like to work on to expand the the supported semantics for QEC researchers. 

1. There is a limitation on how far back you can index measurements that have occurred. This seems fine for something like a surface code memory experiment (which I have as a unit test) but becomes problematic for something like color code memory experiments where, in initialization, you have two sets of measurements that happen on the same qubits and then outside of the loop, you declare a time-based detector. That detector needs to access measurements that exceed the number of qubits being measured.

```
REPEAT 2 {
	MR 0 1 2
}
DETECTOR rec[-3] rec[-6] # you wouldn't be able to access this
```

I imagine if the number of iterations is propagated and perhaps some mechanism to accumulate measurements (say, py.binop.add with ilists) is carefully relaxed on invariance checking this could be supportable.

2. Potentially premature optimization but it might be more space efficient to have some lattice element that represents a RANGE of record indices. `GetItem` naturally complicates things but it would encourage a “lazier” growth of the GlobalRecordState than the current structure.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "bloqade-circuit"
-version = "0.12.0-DEV"
+version = "0.13.0-DEV"
 description = "The software development toolkit for neutral atom arrays."
 readme = "README.md"
 authors = [
--- a/src/bloqade/gemini/post_processing.py
+++ b/src/bloqade/gemini/post_processing.py
@@ -1,0 +1,90 @@
+import typing
+
+import numpy as np
+from kirin import ir
+
+from bloqade.analysis.measure_id import MeasurementIDAnalysis, lattice
+
+T = typing.TypeVar("T")
+
+
+def _has_no_none(value: tuple[T | None, ...]) -> typing.TypeGuard[tuple[T, ...]]:
+    return all(v is not None for v in value)
+
+
+def _post_processing_function(
+    value: lattice.MeasureId,
+) -> typing.Callable[[typing.Sequence[bool]], typing.Any] | None:
+    if isinstance(value, lattice.RawMeasureId):
+
+        def _measure_func(measurements: typing.Sequence[bool]):
+            return bool(measurements[value.idx])
+
+        return _measure_func
+    elif isinstance(value, (lattice.DetectorId, lattice.ObservableId)):
+        measurement_func = _post_processing_function(value.data)
+        if measurement_func is None:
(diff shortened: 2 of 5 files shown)


src/bloqade/stim/passes/soft_flatten.py
```diff
@@ -0,0 +1,92 @@
...
+)
+from kirin.dialects import scf, ilist
+from kirin.rewrite.abc import RewriteResult
+
+# from bloqade.qasm2.passes.fold import AggressiveUnroll
+from bloqade.stim.passes.simplify_ifs import StimSimplifyIfs
+
+
+@dataclass
+class AggressiveUnroll(Pass):
```
cduck: Do we really need another copy of AggressiveUnroll?

src/bloqade/stim/passes/squin_to_stim.py
```diff
@@ -35,35 +36,36 @@ class SquinToStimPass(Pass):
...
+            dialects=mt.dialects, no_raise=self.no_raise
+        ).fixpoint(mt)
 
         # after this the program should be in a state where it is analyzable
         # -------------------------------------------------------------------
 
         mia = MeasurementIDAnalysis(dialects=mt.dialects)
         meas_analysis_frame, _ = mia.run(mt)
+        print("meas_analysis_frame")
+        mt.print(analysis=meas_analysis_frame.entries)
```
weinbe58:
```suggestion
```

test/analysis/address/test_qubit_analysis.py
```diff
@@ -292,3 +292,22 @@ def main():
...
+
+        q = qs[2]
+        return q
+
+    address_analysis = address.AddressAnalysis(main.dialects)
+    frame, _ = address_analysis.run(main)
+    main.print(analysis=frame.entries)
+
+
+test_if_else()
```
weinbe58:
```suggestion
```

Issue #299, opened 2025-05-27
split if is not handling yield values


Roger-luo:
this line is wrong

https://github.com/QuEraComputing/bloqade-circuit/blob/dc25d772cd169152cfa54c6a2c00a98fdec95c2d/src/bloqade/qasm2/rewrite/split_ifs.py#L52

it should not just insert empty `Yield` or `Return` but should split things as is first then run a separate pass to convert non-QASM compatible statements (e.g alias) into QASM2 compatible statements. For example, it is not acting correctly on the following IR where each `if` statement has a yield value

```python
func.func loop_main() -> !py.NoneType {
  ^0(%loop_main_self):
  │         %0 = py.constant.constant 1 : !py.int
  │    %target = qasm2.core.qreg.new(n_qubits=%0) : !py.IList[!py.Qubit, !Any]
  │         %1 = py.constant.constant 0 : !py.int
  │  %target_1 = qasm2.core.qreg.get(reg=%target, idx=%1) : !py.Qubit
  │     %theta = py.constant.constant 0.1 : !py.float
  │      %creg = qasm2.core.creg.new(n_bits=%0) : !py.CReg
  │         %2 = py.constant.constant IList(range(0, 5)) : !py.IList[!py.int, Literal(5,int)]
  │       %ctr = py.constant.constant 0 : !py.int
  │         %3 = py.constant.constant 2 : !py.int
  │         %4 = py.binop.pow(%3, %ctr) : !py.int
  │         %5 = py.binop.mult(%theta, %4) : !py.float
  │   %ancilla = func.invoke prep_resource_state(%5) : !py.Qubit maybe_pure=False
  │         %6 = py.constant.constant 0 : !py.int
  │         %7 = qasm2.core.creg.get(reg=%creg, idx=%6) : !py.Bit
  │         %8 = py.cmp.eq(lhs=%7, rhs=%6) : !py.bool
  │  %target_2 = scf.if %8 {
  │              │ ^1(%36):
  │              │ │             qasm2.uop.CX(ctrl=%ancilla, qarg=%target_1)
  │              │ │       %34 = py.constant.constant 0 : !py.int
  │              │ │       %35 = qasm2.core.creg.get(reg=%creg, idx=%34) : !py.Bit
  │              │ │             qasm2.core.measure(qarg=%target_1, carg=%35)
  │              │ │ %target_7 = py.assign.alias target = %ancilla
  │              │ │             scf.yield %target_7
  │              } else {
[...]

PR #631, opened 2025-11-20
Validation for QASM2 main kernels

### LINKED ISSUE #188: Restrict `qasm2.main` if-statement via validation pass.
use Validation infrastructure in #179 to restrict:

1. If statement doesn't have an else block
2. If statement only has one expression inside the body.

david-pl:
Closes #188  and #299

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "bloqade-circuit"
-version = "0.11.0-DEV"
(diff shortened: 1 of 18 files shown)


david-pl:
> can we fix the docstring error for this PR?

CI is currently set up to check all the docstrings. We were planning a docathon at one point. I'll try to change it so it only checks files with changes in a separate PR.

PR #638, opened 2025-11-25
Bug Fix in Squin To Native rewrites. 

### LINKED ISSUE #633: Issue with rewrites from Squin to Native.
There are multple issues here:

1. TypeInference issue related to `AggressiveUnroll` which might be fixed in the latest version of kirin
2. The `SquinToNative` rewrite is incorrect because it replaces the statements with a call to a library function for rotation gates which creates a problem because the angle is being divided by 2*pi twice because the rule is replacing the gate statement which takes turns as units with a subroutine which takes radians as inputs. 

here is a full example:

```python
from bloqade import squin, native
from bloqade.native.upstream import SquinToNative
from bloqade.rewrite.passes import AggressiveUnroll
from bloqade.test_utils import assert_methods

@squin.kernel
def ghz(angle: float):
    qubits = squin.qalloc(1)
    squin.rz(angle, qubits[0])


@native.kernel
def ghz_native(angle: float):
    qubits = squin.qalloc(1)
    native.rz(angle, qubits[0])

ghz_native_rewrite = SquinToNative().emit(ghz)
AggressiveUnroll(ghz.dialects).fixpoint(ghz_native_rewrite)
ghz_native_rewrite.print()


AggressiveUnroll(ghz_native.dialects).fixpoint(ghz_native)
ghz_native.print()
assert_methods(ghz_native_rewrite, ghz_native)
```

weinbe58:
closes #633 

Two bugs found:

1. bug when rewrite from Squin to Native statements. I need to implement a turns version in the standard library to rewrite gates to. 
2. Bug in `CallGraphPass` that was not applying the rewrite to the method passed into the pass itself.

### DIFF
--- a/src/bloqade/cirq_utils/lowering.py
+++ b/src/bloqade/cirq_utils/lowering.py
@@ -282,8 +282,6 @@ class Squin(lowering.LoweringABC[cirq.Circuit]):
             )
         raise lowering.BuildError(f"Cannot lower {node}")
 
-        # return self.visit_Operation(state, node)
-
     def lower_literal(self, state: lowering.State[cirq.Circuit], value) -> ir.SSAValue:
         raise lowering.BuildError("Literals not supported in cirq circuit")
 
@@ -658,3 +656,10 @@ class Squin(lowering.LoweringABC[cirq.Circuit]):
                 probabilities, controls=control_qarg, targets=target_qarg
             )
         )
+
+    def visit_ResetChannel(
+        self, state: lowering.State[cirq.Circuit], node: cirq.ResetChannel
+    ):
+        qubits = self.lower_qubit_getindices(state, node.qubits)
+        stmt = qubit.stmts.Reset(qubits)
+        return state.current_frame.push(stmt)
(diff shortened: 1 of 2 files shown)


src/bloqade/native/stdlib/broadcast.py
```diff
@@ -21,6 +21,30 @@ def _radian_to_turn(angle: float) -> float:
     return angle / (2 * math.pi)
 
 
+@kernel
+def _rx_turns(angle: float, qubits: ilist.IList[qubit.Qubit, Any]):
```
david-pl: I'm a bit torn with this: on the one hand, it's convenient to rewrite to stdlib. On the other hand, I feel like introducing a private stdlib function just for the rewrite is a bit of an overkill. Everything is in units of `turns` and broadcasted at the statement level, so why not just rewrite statements to statements directly? Sure, you have to basically duplicate the code from the stdlib and inject it into the IR, but introducing these functions is also a bit redundant.

Generally, one disadvantage of using rewrites to stdlib is that you are introducing a bunch of `Invoke`s in what is probably a flattened kernel (e.g. coming from `gemini.logical`). So you'll have to flatten after the rewrite again. This would also be avoided when rewriting to statements directly.

david-pl (approved):
Looks good, but I do have some questions. They don't necessarily require changes in this PR though.

PR #640, opened 2025-11-26
Properly implement fidelity analysis with SCF

### LINKED ISSUE #592: Proper implementation of the fidelity analysis
The current version of the fidelity analysis is essentially broken and not very useful. We need to refactor it in order to properly work. This was previously blocked by the fact that we couldn't run the qubit address analysis on scf expression, but with #563 that should be fixed.

david-pl:
Closes #592 .

We are now mostly re-using the `AddressAnalysis` for walking the IR. Except for `IfElse` statements, where we need to take care when merging results from branches.
To do this, I had to make a small change to `AddressAnalysis`, but that shouldn't affect anything else.

Probably needs more tests, but basics seem to work quite well.

Also, we might need some wrapper around running this analysis to have a decent UX when trying to study the fidelity of a given program.

@weinbe58 I'm also not sure if I'm being to strict about expecting an `AddressReg` for `qubtis: IList[Qubit]` arguments.

### DIFF
--- a/src/bloqade/analysis/address/impls.py
+++ b/src/bloqade/analysis/address/impls.py
@@ -366,16 +366,22 @@ class Scf(interp.MethodTable):
         if iter_type is None:
             return interp_.eval_fallback(frame, stmt)
 
+        body_values = {}
         for value in iterable:
             with interp_.new_frame(stmt, has_parent_access=True) as body_frame:
                 loop_vars = interp_.frame_call_region(
                     body_frame, stmt, stmt.body, value, *loop_vars
                 )
 
+            for ssa, val in body_frame.entries.items():
+                body_values[ssa] = body_values.setdefault(ssa, val).join(val)
+
             if loop_vars is None:
                 loop_vars = ()
 
             elif isinstance(loop_vars, interp.ReturnValue):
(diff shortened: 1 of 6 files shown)


src/bloqade/analysis/fidelity/impls.py
```diff
@@ -0,0 +1,62 @@
...
+        # run both branches
+        with interp_.new_frame(stmt, has_parent_access=True) as then_frame:
+            # NOTE: reset fidelities before stepping into the then-body
+            interp_.reset_fidelities()
+
+            interp_.frame_call_region(
+                then_frame,
+                stmt,
+                stmt.then_body,
+                *(interp_.lattice.bottom() for _ in range(len(stmt.args))),
```
weinbe58: two things:
1. IfElse only takes one argument, the condition
2. The block only takes one argument which is the condition.

Because of this you should pass a lattice element with Constant `True` for the then block and `False` for the else block.

weinbe58 (changes requested):
Overall looks good, just one comment on the implementation of `IfElse`. Also I would add a test to make sure that you execute only one branch if the branch value is known at compile time.

PR #641, opened 2025-12-01
TerminalLogicalMeasurement Validation

johnzl-777:
This is what I think should be sufficient for TerminalLogicalMeasurement Validation. 

I enforce the following conditions:
1. A program should only ever have ONE TerminalLogicalMeasurement
2. all spawned qubits should be consumed in that one terminal measurement. There should be no dangling qubits.

There are some additions that come to mind - ranging from immediately relevant to some things that are further out in terms of infrastructure:
- I had to make a separate validation pass as opposed to tacking things onto the current logical validation. I think because of this "logical validation" should be renamed to something a bit more precise.
- AddressAnalysis already counts the number of (unique) qubits spawned but there's no way to access that information directly. It would be nice if there was some way to attach that to the final frame returned at the end of the analysis. Currently I tally up unique qubit ids and compare that against the number of things fed into the terminal measurement

### DIFF
--- a/src/bloqade/gemini/analysis/measurement_validation/impls.py
+++ b/src/bloqade/gemini/analysis/measurement_validation/impls.py
@@ -65,7 +65,7 @@ class __GeminiLogicalMeasurementValidation(_interp.MethodTable):
                 stmt,
                 ir.ValidationError(
                     stmt,
-                    "Measurement ID Analysis failed to produce the necessary results needed for vaolidation.",
+                    "Measurement ID Analysis failed to produce the necessary results needed for validation.",
                 ),
             )
             return (interp.lattice.bottom(),)
--- a/test/gemini/test_logical_validation.py
+++ b/test/gemini/test_logical_validation.py
@@ -124,7 +124,9 @@ def test_terminal_measurement():
 
     main.print()
 
-    @gemini.logical.kernel(no_raise=False, aggressive_unroll=True, typeinfer=True)
+    @gemini.logical.kernel(
+        verify=False, no_raise=False, aggressive_unroll=True, typeinfer=True
+    )
     def not_all_qubits_consumed():
         qs = squin.qalloc(3)
         sub_qs = qs[0:2]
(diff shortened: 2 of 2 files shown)


src/bloqade/gemini/analysis/measurement_validation/impls.py
```diff
@@ -0,0 +1,129 @@
...
+
+        # should just be one MeasureIDTuple
+        measure_lattice_element = measurement_analysis_results.get_values(stmt.results)[
+            0
+        ]
+
+        # Figure out the total number of qubits spawned, keeping in mind that if a user
+        # "shuffles" the qubits (puts them in a new container, splits one off from a container type, etc.)
+        # it should be accounted for. This would be much cleaner if there was a way to propagate the
+        # final qubit count saved in the actual interpreter for address analysis...
```
david-pl: I don't understand this comment: you're running the address analysis, right? Why can't you get the `AddressAnalysis.qubit_count` value at some point?

johnzl-777: I think I talked about this with Phil because I thought of this too, that unique qubit counter is attached to the analysis itself, not the frame - which is the only thing I have access to in validation. 

I don't think there's a nice way to affix that to the final frame? Could be wrong though

johnzl-777: Just putting this down for posterity, @david-pl pointed out I can just manually run the address analysis interpreter in the validation pass and route the info I need instead of relying on the ValidationPass auto-run infrastructure (:

src/bloqade/gemini/analysis/measurement_validation/impls.py
```diff
@@ -0,0 +1,88 @@
...
+
+        measurement_analysis_results = interp.measurement_analysis_results
+        total_qubits_allocated = interp.unique_qubits_allocated
+
+        # could make these proper exceptions but would be tricky to communicate to user
+        # without revealing under-the-hood details
+        measure_lattice_element = measurement_analysis_results.get_values(stmt.results)
+        assert len(measure_lattice_element) == 1
+        measure_lattice_element = measure_lattice_element[0]
+        assert isinstance(measure_lattice_element, MeasureIdTuple)
```
weinbe58:
```suggestion
        if not isinstance(measure_lattice_element, MeasureIdTuple):
            # add validation error here because analysis failed to figure out the measurements. 
            ...
```

weinbe58 (changes requested):
overall looks good 🚀  just a few comments to address

weinbe58 (approved):
LGTM

Issue #540, opened 2025-10-10
rewrite QASM2 dialect to SQUIN


Roger-luo:
similar to STIM but for QASM2, right now QASM2 only lowers from AST to the QASM2 dialect.

PR #644, opened 2025-12-04
QASM2 to squin

johnzl-777:
This is my implementation of #540 , there is one thing that I've held off on after talking with @weinbe58 that's a bit tricky: handling measurements. My understanding is we don't need it for now but in the future it will be relevant. 

While 95% of qasm2 maps very nicely to existing squin infrastructure (with the exception of some of the less common 2 and 3 qubit gates I had to skip - unless we want to do some implicit conversion for them) you run into problems trying to map the qasm2 measurement semantics. 

Take the following examples:

1. A user wants to measure out all the qubits to the entirety of the declared register. Your qasm2 looks something like this
```
qreg q[5]
creg c[5]
measure q -> c
```
You could potentially just have creg be an empty ilist and then when the measure happens produce the ilist of measurements - making sure that all subsequent references point to the new ilist. This seems acceptable but it gets tricky if a user wants to do something like 

2. a user only wants to measure a single qubit and save that to a _slot_ in the classical register. This looks like

```
qreg q[5]
creg c[5]
measure q[0] -> c[0]
```
Now the classical register somehow has to already be `get_item`-able which is tricky considering there's no notion of a fixed-length list. Things get even trickier if a user wants to save at a different index:

```
measure q[1] -> c[3]
```

P.S. The sub kernel rewrite seems "off" to me, I'm sure there must be a better way to do it but I'm not aware of what the proper API for it is 😅

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -43,7 +43,7 @@ jobs:
           token: ${{ secrets.CODECOV_TOKEN }} # required
       - name: Archive code coverage results
         if: matrix.python-version == '3.12'
-        uses: actions/upload-artifact@v5
+        uses: actions/upload-artifact@v6
         with:
           name: code-coverage-report
           path: coverage.xml
@@ -55,7 +55,7 @@ jobs:
     steps:
         - uses: actions/checkout@v6
         - name: download covearge
-          uses: actions/download-artifact@v6
+          uses: actions/download-artifact@v7
           with:
             name: code-coverage-report
         - name: check coverage
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -19,7 +19,7 @@ jobs:
     - name: Build distribution 📦
       run: uv build
     - name: Store the distribution packages
-      uses: actions/upload-artifact@v5
+      uses: actions/upload-artifact@v6
       with:
         name: python-package-distributions
         path: dist/
@@ -39,7 +39,7 @@ jobs:
 
     steps:
     - name: Download all the dists
-      uses: actions/download-artifact@v6
+      uses: actions/download-artifact@v7
       with:
         name: python-package-distributions
         path: dist/
@@ -60,7 +60,7 @@ jobs:
 
     steps:
     - name: Download all the dists
-      uses: actions/download-artifact@v6
+      uses: actions/download-artifact@v7
       with:
         name: python-package-distributions
         path: dist/


johnzl-777:
Latest round of feedback: use dictionary style mapping (faster than pattern matching) + can use pre-existing QASM2 to py expr rewriterule

src/bloqade/squin/rewrite/qasm2/noise_to_squin.py
```diff
@@ -0,0 +1,80 @@
+from kirin import ir
+from kirin.dialects import func
+from kirin.rewrite.abc import RewriteRule, RewriteResult
+
+from bloqade import squin
+from bloqade.qasm2.dialects.noise import stmts as noise_stmts
+
+from .util import num_to_py_constant
```
david-pl: Do we have to create a `util` file for a single function used in a single file? 😅

src/bloqade/squin/rewrite/qasm2/core_to_squin.py
```diff
@@ -0,0 +1,38 @@
...
+
+        if isinstance(node, core_stmts.QRegNew):
+            args = (node.n_qubits,)
+        elif isinstance(node, core_stmts.Reset):
+            args = (node.qarg,)
+        else:
+            return RewriteResult()
+
+        new_stmt = func.Invoke(
+            callee=CORE_TO_SQUIN_MAP[type(node)],
```
david-pl: General question: Why do you rewrite to stdlib calls rather than statements?

johnzl-777: In my head it felt nicer to default to the stdlib versions because the output to the user would be on par with what they would see if they constructed the squin kernel through the interface we provide. Figured it makes debugging a bit nicer.

That being said, I don't have a strong preference it _has_ to be that way and if it ends up being the case we prefer the "unrolled"/flat form as the output I can easily do that (:

src/bloqade/squin/rewrite/qasm2/qasm2_modified_to_squin.py
```diff
@@ -0,0 +1,83 @@
...
+            new_args = (pi_over_2_stmt.result, node.phi, node.lam, node.qarg)
+        elif isinstance(node, (uop_stmts.UGate, parallel.UGate, glob.UGate)):
+            new_args = (
+                node.theta,
+                node.phi,
+                node.lam,
+                (
+                    node.qarg
+                    if hasattr(node, "qarg")
+                    else (node.registers if hasattr(node, "registers") else node.qargs)
```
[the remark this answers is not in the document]
johnzl-777: Good catch! I've gone ahead and corrected the rewrite logic to properly turn a glob.U into multiple broadcast.u3 instances

test/squin/passes/test_qasm2_to_squin.py
```diff
@@ -0,0 +1,454 @@
...
+        Chain(
+            QASM2ToPyRule(),
+            qasm2_rules.QASM2DirectToSquin(),
+            qasm2_rules.QASM2ModifiedToSquin(),
+        )
+    ).rewrite(global_parallel_program.code)
+    AggressiveUnroll(dialects=global_parallel_program.dialects).fixpoint(
+        global_parallel_program
+    )
+
```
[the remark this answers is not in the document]
johnzl-777: Should be fixed now 👍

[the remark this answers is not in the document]
david-pl: Hm, I still get

```python
func.func @global_program() -> !Bottom {
```

This might be a kirin issue, but I'm not sure. Anyway, I don't think this should be a blocker here.

src/bloqade/squin/rewrite/qasm2/qasm2_modified_to_squin.py
```diff
@@ -81,3 +78,21 @@ def rewrite_Statement(self, node: ir.Statement) -> RewriteResult:
...
+
+    def rewrite_glob_UGate(self, node: glob.UGate) -> RewriteResult:
+
+        # assume that QASM2DirectToSquin has already run, which converts QRegNew to qubit.new in the squin dialect.
+        # Then you're left with an ilist of ilists of qubit.news, for which a new squin.broadcast.u3 should be created.
+
+        ilist_of_registers = node.registers.owner
+        assert isinstance(ilist_of_registers, ilist.New)
+
+        for register_ilist in ilist_of_registers.values:
```
david-pl: If we want to be really fancy, we could do this with an `ilist_of_registers.map(squin.broadcast.u3)` or so, but this is also fine.

PR #653, opened 2025-12-05
Fix state initialization validation in gemini logical

david-pl:
Previously, we'd only allow any single non-Clifford gate to be applied to a set of qubits. That means, you couldn't e.g. apply an `Rx` to the first and an `Ry` to the second qubit, since the latter would have already counted as the second gate and it's non-Clifford.

This fixes this by actually checking the first gates per qubit address.

### DIFF
--- a/src/bloqade/gemini/dialects/logical/_interface.py
+++ b/src/bloqade/gemini/dialects/logical/_interface.py
@@ -1,4 +1,4 @@
-from typing import TypeVar
+from typing import Any, TypeVar
 
 from kirin import lowering
 from kirin.dialects import ilist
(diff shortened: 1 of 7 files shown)


src/bloqade/gemini/analysis/logical_validation/analysis.py
```diff
@@ -1,27 +1,50 @@
...
 
         return tuple(self.lattice.bottom() for _ in range(len(node.results)))
 
+    def check_first_gate(self, qubits: ir.SSAValue) -> bool:
+        address = self.addr_frame.get(qubits)
+
+        if isinstance(address, AddressQubit):
+            is_first = self.first_gates.get(address.data, True)
+            self.first_gates[address.data] = False
+            return is_first
```
weinbe58: based on the semantics this branch shuold never execute since it will always be an Ilist of qubits.

PR #672, opened 2026-01-13
use bloqade-decoders as source of annotation dialect + relevant types

johnzl-777:


pyproject.toml
```diff
@@ -11,6 +11,7 @@ authors = [
 ]
 requires-python = ">=3.10"
 dependencies = [
+    "bloqade-decoders",
```
david-pl: This should have a compat bound, e.g. `~=0.1.0`.

PR #688, opened 2026-01-28
Support Stim tags and export Stim non-Clifford gates

ChenZhao44:
This PR includes changes to support stim tags for Stim circuit analysis. It also exports the non-Clifford gates explicitly in the dialect, allowing users to run potential simulations with Tsim.

To support the Stim tag, I introduced a base class for all Stim native statements, including gate, noise, collapse, and annotation. Those instructions are inherent from the `StimStatement` base class that includes the tag.

One thing I would like to mention is that the tag now is typed as `tag: str = info.attribute(types.String, default=None)` instead of `str | None`. I couldn't figure out how to make the `types.UnionType` work for attributes. Probably it is a bug of the `UnionType`. Therefore, the current type will result in an error when the user explicitly sets `tag=None`. For example,
```python
from bloqade import stim

@stim.main
def stim_kernel():
    stim.x(tag=None, targets=(0, 1))
```
will not pass the compilation.


johnzl-777:
I talked a bit with @rafaelha in person on this + review, the utility doesn't quite make sense anymore considering 

1. The `stim` dialect itself is not meant to be something users interact with
2. For tsim compatibility, the necessary subset of stim statements with tags is more than sufficient
3. If, it is the case in the future a consumer will need tagged statements, it probably won't require such a degree of genericism. 

If it is the case this flexibility is required in the future, I'd be more than happy to bring it back but closing this PR for now (:

Issue #693, opened 2026-01-29
Unexpected behaviour in heuristic noise model


AlexSchuckert:
This line in bloqade->cirq_utils->noise->transform.py->transform_circuit is problematic:

`compressed_circuit = cirq.merge_single_qubit_moments_to_phxz(interleaved_circuit)`

It merges single qubit gates, which might not be wanted by the user (e.g. when simulating randomized benchmarking)

I propose to change it to

`if not to_native_gateset:
compressed_circuit = cirq.merge_single_qubit_moments_to_phxz(interleaved_circuit)
else:
compressed_circuit = interleaved_circuit`

PR #694, opened 2026-01-30
Remove single gate merging from noise model

### LINKED ISSUE #659: Noise model doesn't work with identity operations
Running the following snippet of code (to explore the interplay between doing local gates, vs. global gates and inverting a subset). However, even with `to_native_gateset=False` the transformer still cuts out the identity qubits. And, I have to write the identity as a `PhXZ` gate otherwise there is an error.

```python
sqrt_Y = cirq.PhasedXZGate(axis_phase_exponent=0.5,x_exponent=0.5,z_exponent=0)
sqrt_Y_dag = cirq.PhasedXZGate(axis_phase_exponent=0.5,x_exponent=-0.5,z_exponent=0)
identity = cirq.PhasedXZGate(axis_phase_exponent=0.0,x_exponent=0.0,z_exponent=0)

qubits = cirq.LineQubit.range(8)
fidelity_local = []
fidelity_global = []
for k in range(1,8):
    circuit_global = cirq.Circuit.from_moments(
        cirq.Moment([identity.on(q) for q in qubits]),
        cirq.Moment([sqrt_Y.on(q) for q in qubits]),
        cirq.Moment([sqrt_Y_dag.on(q) for q in qubits[k::]]))
    circuit_local = cirq.Circuit.from_moments(
        cirq.Moment([identity.on(q) for q in qubits]),
        cirq.Moment([sqrt_Y.on(q) for q in qubits[0:k]]))
    noisy_circuit_local = transform_circuit(circuit_local, model=model,to_native_gateset=False)
    noisy_circuit_global = transform_circuit(circuit_global, model=model,to_native_gateset=False)
    
    # Simulate noisy circuits
    clean = simulator_dm.simulate(circuit_local).final_density_matrix
    noisy = simulator_dm.simulate(noisy_circuit_local).final_density_matrix
    fidelity_local.append(np.trace(clean @ noisy).real)
    
    clean = simulator_dm.simulate(circuit_global).final_density_matrix
    noisy = simulator_dm.simulate(noisy_circuit_global).final_density_matrix
    fidelity_global.append(np.trace(clean @ noisy).real)
plt.plot(fidelity_local,label='Local')
plt.plot(fidelity_global,label='Global')
plt.legend()
plt.show()
```

david-pl:
Closes #659 and #693

### DIFF
--- a/src/bloqade/cirq_utils/noise/model.py
+++ b/src/bloqade/cirq_utils/noise/model.py
@@ -442,14 +442,11 @@ class GeminiOneZoneNoiseModel(GeminiNoiseModelABC):
         interleaved_circuit = cirq.Circuit.from_moments(*interleaved_moments)
 
         # Combine subsequent 1Q gates
-        compressed_circuit = cirq.merge_single_qubit_moments_to_phxz(
-            interleaved_circuit
-        )
         if self.parallelize_circuit:
-            compressed_circuit = parallelize(compressed_circuit)
+            interleaved_circuit = parallelize(interleaved_circuit)
 
         return self._noisy_moments_impl_moment(
-            compressed_circuit.moments, system_qubits
+            interleaved_circuit.moments, system_qubits
         )
 
 


tcochran-quera (commented):
The edge case that this misses is if you have a qubit that doesn't participate in a CZ gate for multiple layers, but might have multiple 1Q gates applied during that time (for example, after the final CZ gate and before the next one)... you probably want to automatically combine those single qubit gates.

For example, if you have this circuit:

```
0: ───X───────────────X───
      │               │
1: ───@───@───X───@───@───
          │   │   │
2: ───────X───@───X───────
```

This translates to the hardware like:

```
0: ───H───@───H───────────────────────H───@───H───
          │                               │
1: ───────@───────@───H───@───H───@───────@───────
                  │       │       │
2: ───────────H───@───H───@───H───@───H───────────
```

but we can get ride of some Hadamards like:

```
0: ───H───@───────────────────────────────@───H───
          │                               │
1: ───────@───────@───H───@───H───@───────@───────
                  │       │       │
2: ───────────H───@───H───@───H───@───H───────────
```

which wouldn't be caught if we remove line 445.

So, maybe we keep the `merge_single_qubit_moments` only `if parallelize_circuit or to_target_gateset`? @AlexSchuckert does that work for you?

PR #700, opened 2026-02-04
Improve docstring CI

david-pl:
This changes the docstring CI to only check the docstrings of files that were changed within a PR. Also, I simplified it a bit so it only checks for missing docstrings, not specific styling etc.

So now, whenever you change a file, the pre-commit and CI will make you add docstrings throughout the entire file. Hopefully this will allow us to incrementally improve the docstring situation.

Some open questions:
* Right now, I'm skipping the rule `D100`, which checks for missing docstrings in public modules, which we don't have anywhere. Do we want to have those? This would certainly make the automatically built API reference look nicer.
* Do we want to enforce some specific style?


Roger-luo (approved):

Roger-luo:
Sorry I must miss this PR review somehow, GTM

PR #705, opened 2026-02-13
Bug Fixes in Native standard library. 

weinbe58:
Fixing bugs in native standard library implementation and adding tests to compare native and squin. 

Note that I fixed the tests and confirmed that the new tests should match the actual computation of `sqrt_y.dot([1, 0])` and `sqrt_y.T.conj().dot([1, 0])`


david-pl (approved):
I think this should be good now.

One general question: these decompositions are equivalent up to single-qubit phase. Are we sure those phases will never come back to haunt us? 😅

PR #708, opened 2026-02-18
Implement Gemini logical device with post-processing API

### LINKED ISSUE #645: Gemini Logical Device API for Post processing.
This Issue supersedes #643. 

In order to support the QEC use cases we need to have some way of specifying the post processing of the physical measurements from the logical execution. Here I propose that we just pass the post processing function when submitting the tasks to the QPU that way we do not need to explicitly analyze the users kernel to figure out what is post processing and what is QPU logic.

# Add a callback function to future

In many cases with future objects you can specify a callback function that should be executed on the results of the future. We can barrow that logic here at the task level:

```python
post_processed_results = (
    device.task(quantum_kernel).run_async(*args, **kwargs)
    .callback(post_processing).get_result()
)
```

in terms of the signatures here we have:

```python
class Device:
    def task(kernel: ir.Method[Args, Return]) -> Task[Args, Return]:
        ...


class Task(Generic[Args, Return]):
    def run_async(self, *args: Args.args, **kwargs: Args.kwargs) -> Future[Return]:
        ...


class Future(Generic[Return]):
    def callback(self, callback: ir.Method[[Return],T]) -> Future[T]:
        ...

    def get_result() -> Return:
        ...

```

david-pl:
Closes #645.

Requires #704 to merged first (currently this is merged into the branch).

For now, we just run a simple PyQrack simulation for the logical qubit number and assume no errors, i.e. the measurement result for each logical qubit is just included `num_physical_qubits` times.


weinbe58:
this branch is stale can I close this since its moved to bloqade-internal?

david-pl:
Yep

PR #711, opened 2026-02-23
feat: qasm3 module

yitchen-tim:
## Add OpenQASM 3.0 support

This PR adds a complete OpenQASM 3.0 workflow to bloqade-circuit: parsing, IR construction, emission, and conversion to squin.

### What's included

**QASM3 dialect definitions** (`src/bloqade/qasm3/dialects/`)
- `core` — quantum/classical register operations (`QRegNew`, `BitRegNew`, `QRegGet`, `BitRegGet`, `Measure`, `Reset`, `Barrier`)
- `uop` — quantum gates (`H`, `X`, `Y`, `Z`, `S`, `T`, `RX`, `RY`, `RZ`, `CX`, `CY`, `CZ`, `UGate`)
- `expr` — arithmetic expressions, constants, and `GateFunction` (custom gate subroutine IR node)

**Python decorator interface** (`src/bloqade/qasm3/groups.py`, `_wrappers.py`)
- `@qasm3.gate` for defining custom gates
- `@qasm3.main` for defining programs
- Wrapper functions (`qasm3.h`, `qasm3.cx`, `qasm3.qreg`, `qasm3.measure`, etc.) that lower Python calls to IR statements

**Parsing** (`src/bloqade/qasm3/parse/`)
- `qasm3.loads(source)` and `qasm3.loadfile(path)` parse OpenQASM 3.0 text into `ir.Method`
- Uses `openqasm3` parser for AST, then `QASM3Lowering` visitor to build kirin IR
- Supports qubit/bit declarations, gate applications, custom gate definitions, measurements, resets, barriers, and arithmetic expressions

**Emission** (`src/bloqade/qasm3/emit/`)
- `QASM3Emitter().emit(method)` converts IR back to valid OpenQASM 3.0 text
- Uses kirin's abstract interpretation (`EmitABC`) with dialect-registered method tables
- Handles main program body (`EmitQASM3Main`) and custom gate definitions (`EmitQASM3Gate`)

**QASM3 → Squin conversion** (`src/bloqade/squin/passes/qasm3_to_squin.py`, `qasm3_gate_func_to_squin.py`)
- `QASM3ToSquin` pass converts QASM3 IR to squin dialect
- Follows the same `CallGraphPass` pattern as `QASM2ToSquin` — converts custom gate bodies across the call graph
- Rewrite rules: `QASM3DirectToSquin` (1:1 gate mappings) and `QASM3ModifiedToSquin` (argument reordering for rotation gates and UGate)

**Dependencies**
- Added `openqasm3[all]~=1.0.1` and `lark>=1.2.2` as optional `[qasm3]` dependencies

### Control Flow
1. **`for` loops** (`ForInLoop`) — iteration over ranges/sets
2. **`while` loops** (`WhileLoop`) — conditional looping
3. **`if`/`else` branching** (`BranchingStatement`) — classical conditional execution
4. **`switch` statements** (`SwitchStatement`) — multi-way branching
5. **`break` / `continue`** — loop control flow

### Subroutines & Scoping
6. **Subroutine definitions** (`SubroutineDefinition`) — classical `def` functions (gate definitions *are* supported)
7. **`return` in subroutines** — returning values from classical functions
8. **`extern` declarations** — opaque references to external classical functions

### Types & Declarations
9. **`angle` type** — fixed-point angle representation
10. **`const` declarations** — compile-time constant variables
11. **Type casting** (`cast`) — explicit type conversions between classical types
12. **`duration` type** — timing durations
13. **Array types** — `array[int[32], 10]` and multi-dimensional arrays

### Expressions & Operators
14. **Comparison operators** — `==`, `!=`, `<`, `>`, `<=`, `>=`
15. **Logical operators** — `&&`, `||`, `!` (logical NOT)
16. **Bit-shift operators** — `<<`, `>>`
17. **Bitwise AND/OR/XOR** — `&`, `|`, `^`
18. **Ternary operator** — `condition ? a : b`
19. **Concatenation expressions** — register concatenation
20. **Additional math functions** — `arcsin`, `arccos`, `arctan`, `ceiling`, `floor`, `popcount`, `rotl`, `rotr`
21. **Additional constants** — `tau` (τ), `euler` (ℇ) (only `pi` is supported)

### Quantum Features
22. **Gate modifiers** — `ctrl @`, `negctrl @`, `inv @`, `pow(k) @`
23. **Multi-dimensional qubit/bit indexing** — only simple `q[i]` is supported
24. **Physical qubits** — `$0`, `$1` hardware qubit references
25. **Register aliasing** (`let`) — creating aliases for register slices

### Timing & Pulses
26. **`delay` statements** — inserting timing delays
27. **`stretch` type** — flexible

### Directives & Metadata
28. **`pragma`** — vendor-specific compiler directives (`#pragma ...`)
29. **Annotations** — metadata annotations on statements (`@annotation`)

### DIFF
--- a/src/bloqade/squin/rewrite/qasm3/qasm3_expr_to_py.py
+++ b/src/bloqade/squin/rewrite/qasm3/qasm3_expr_to_py.py
@@ -50,6 +50,12 @@ class QASM3ExprToPy(RewriteRule):
         elif isinstance(node, (expr.Sin, expr.Cos, expr.Tan, expr.Exp, expr.Sqrt)):
             node.replace_by(self.UNARY_OPS[type(node)](x=node.value))
             return RewriteResult(has_done_something=True)
+        elif isinstance(node, expr.Log):
+            raise NotImplementedError(
+                "natural log (ln) is not yet supported in the QASM3-to-squin"
+                " rewrite pass because the kirin math dialect does not provide"
+                " a log/ln operation"
+            )
         elif isinstance(
             node, (expr.Add, expr.Sub, expr.Mul, expr.Div, expr.Pow, expr.Mod)
         ):
--- a/test/squin/passes/test_qasm3_to_squin.py
+++ b/test/squin/passes/test_qasm3_to_squin.py
@@ -315,3 +315,15 @@ def test_qasm3_expr_to_py_bitnot():
         "gate myg(a) q { rx(~a) q; }\n"
         "qubit[1] q;\nmyg(3) q[0];\n"
     )
+
+
+def test_qasm3_expr_to_py_log_unsupported():
+    """QASM3ExprToPy raises NotImplementedError for log (ln) expression."""
+    import pytest
+
+    with pytest.raises(NotImplementedError, match="natural log"):
+        _load_and_convert(
+            'OPENQASM 3.0;\ninclude "stdgates.inc";\n'
+            "gate myg(a) q { rx(log(a)) q; }\n"
+            "qubit[1] q;\nmyg(0.5) q[0];\n"
+        )


yitchen-tim:
Hi @david-pl thanks for the comments. I have made updates to address the comments. 

I got the linting done with pre-commit. The linting update the code beyond my changes, hopefully that's okay. When I ran unit tests locally, there are some failed tests that is not related to this PR. I didn't try to fix them. 

> Another thing to keep in mind: it seems to me that quite a bit of functionality could be shared between qasm2 and qasm3 (e.g. the expr dialect, types). However, this might need some breaking changes in qasm2. Besides, I feel like this is beyond the scope of this PR. For now, let's just keep thing s separate, but something to keep in mind for later.

I agree keeping them separate makes sense for now. Given that QASM2 has stopped developing while QASM3 is still evolving, I would think the development should center around QASM3. QASM2 should support strictly a subset of QASM3, so I can see some de-duplicate we can do. For example, QASM3 becomes a combination of QASM2 dialects + few new dialect. But right now, the QASM2 dialect isn't defined purely with the spec. It also intertwines with features like schedule, parallel and AddressAnalysis. It's a bit not obvious for me what is a clean way to inherit dialects from QASM2. 

> Do you mind sharing what features specifically you still plan on adding? Specifically, I'm wondering whether you plan on adding control flow?

These are what I have in mind. I might not necessarily do all of them, but they are what I can think of for now. 
- Support control flow
- Add more gates
- Handle include file (gates defined in include file can be used in main)
- Support pragma
- Support Box
- Barrier be used in @qasm3.gate (In qasm3, I put it in core. In qasm2, it’s in uop but I feel it’s not really a unitary. May need a better place for it.)
- Noise
- Parallel dialect
- Pyrack simulations

src/bloqade/cirq_utils/emit/noise.py
```diff
@@ -12,7 +12,7 @@ class __EmitCirqNoiseMethods(MethodTable):
     two_qubit_paulis = (
         "IX",
         "IY",
-        "IZ",
+        "IS",
```
david-pl: Oh no, the `typos` pre-commit hook is causing some issues here it seems.

src/bloqade/qasm3/emit/main.py
```diff
@@ -0,0 +1,126 @@
...
+
+from .base import EmitQASM3Base, EmitQASM3Frame
+
+
+@dataclass
+class EmitQASM3Main(EmitQASM3Base):
+    keys = ("emit.qasm3.main", "emit.qasm3.gate")
+    dialects: ir.DialectGroup
+
+    def initialize(self) -> Self:
```
david-pl: I think both `initialize` and `eval_fallback` here can go since they have the same definition as `EmitQASM3Base`. Just a nit pick though.

src/bloqade/qasm3/emit/main.py
```diff
@@ -0,0 +1,126 @@
...
+
+@func.dialect.register(key="emit.qasm3.main")
+class Func(interp.MethodTable):
+
+    @interp.impl(func.Invoke)
+    def invoke(self, emit: EmitQASM3Main, frame: EmitQASM3Frame, stmt: func.Invoke):
+        callee = stmt.callee
+        gate_name = callee.sym_name
+
+        # After QASM3ToSquin, QRegNew may be rewritten to func.invoke qalloc.
```
david-pl: You shouldn't try to emit QASM3 from a squin kernel. If you want to have that capability, the workflow should be to use a `Pass` that rewrites from squin to QASM3 first. Do you have a use-case for this in mind?

This code path here can go, I think.

src/bloqade/qasm3/dialects/expr/_emit.py
```diff
@@ -0,0 +1,93 @@
...
+        rhs = frame.get(stmt.rhs)
+        return (f"({lhs} * {rhs})",)
+
+    @interp.impl(stmts.Div)
+    def emit_div(self, emit: EmitQASM3Base, frame: EmitQASM3Frame, stmt: stmts.Div):
+        lhs = frame.get(stmt.lhs)
+        rhs = frame.get(stmt.rhs)
+        return (f"({lhs} / {rhs})",)
+
+    @interp.impl(stmts.GateFunction)
```
david-pl: This method causes the tests to fail sometimes: it randomly picks either this impl or the (correct) one in `emit.py`.

As a general rule: never register two methods for the same statement under the same key.
I'm actually surprised this doesn't error.

To fix the build just delete this implementation here.

Issue #403, opened 2025-07-24
Native Squin operator equivilant to `cirq.PhaseZXGate`.


weinbe58:
`cirq.PhaseZXGate` is the single-qubit gate used in the default CZ gate set transpiler in Cirq. The problem is that when you lower a Cirq Circuit it will take these gates and break it into three separate applies:

```python
from bloqade.squin.cirq import load_circuit
import cirq


circ = cirq.Circuit()

circ.append(cirq.PhasedXZGate(axis_phase_exponent=0.5, x_exponent=0.5, z_exponent=0.5).on(cirq.GridQubit(0, 0)))

kernel = load_circuit(circ)

kernel.print()
```

gives:

```mlir
func.func main() -> !py.NoneType {
  ^0(%main_self):
  │  %0 = py.constant.constant 1 : !py.int
  │  %1 = squin.qubit.new(n_qubits=%0) : !py.IList[!py.Qubit, !Any]
  │  %2 = py.constant.constant 1.5707963267948966 : !py.float
  │  %3 = squin.op.shift_op(theta=%2) : !py.Op
  │  %4 = py.constant.constant 0 : !py.int
  │  %5 = py.indexing.getitem(%1, %4) : !Any
  │  %6 = py.ilist.new(values=(%5)){elem_type=!py.Qubit} : !py.IList[!py.Qubit, Literal(1,int)]
  │       squin.qubit.apply(operator=%3, qubits=%6)
  │  %7 = squin.op.x() : !py.Op
  │  %8 = py.constant.constant 1.5707963267948966 : !py.float
  │  %9 = squin.op.rot(axis=%7, angle=%8) : !py.Op
  │ %10 = py.constant.constant 0 : !py.int
  │ %11 = py.indexing.getitem(%1, %10) : !Any
  │ %12 = py.ilist.new(values=(%11)){elem_type=!py.Qubit} : !py.IList[!py.Qubit, Literal(1,int)]
  │       squin.qubit.apply(operator=%9, qubits=%12)
  │ %13 = squin.op.z() : !py.Op
  │ %14 = py.constant.constant 0 : !py.int
  │ %15 = py.indexing.getitem(%1, %14) : !Any
  │ %16 = py.ilist.new(values=(%15)){elem_type=!py.Qubit} : !py.IList[!py.Qubit, Literal(1,int)]
  │       squin.qubit.apply(operator=%13, qubits=%16)
  │ %17 = func.const.none() : !py.NoneType
[...]

PR #722, opened 2026-03-05
Add PhasedXZ Gate to Squin

Jason-Ludmir:
See issue #403

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "bloqade-circuit"
-version = "0.13.0-DEV"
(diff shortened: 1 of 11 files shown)


src/bloqade/cirq_utils/lowering.py
```diff
@@ -190,7 +189,7 @@ class Squin(lowering.LoweringABC[cirq.Circuit]):
     two_qubit_paulis = (
         "IX",
         "IY",
-        "IZ",
+        "IS",
```
weinbe58: I don't know why this is being changed?

PR #724, opened 2026-03-09
Add qubit count to Gemini Logical mode validation. 

### LINKED ISSUE #717: Gemini logical Validation does not validate number of qubit allocations.
Currently the following compiles without an issue:

```python
from bloqade.gemini import logical

@logical.kernel(aggressive_unroll=True, verify=True)
def main()->tuple[list[Detector], list[Observable]]:
    reg = qubit.qalloc(12)
    squin.h(reg[0])
    squin.cx(reg[0], reg[1])
```

which should not be the case at the moment.

weinbe58:
closes #717

### DIFF
--- a/src/bloqade/gemini/analysis/logical_validation/impls.py
+++ b/src/bloqade/gemini/analysis/logical_validation/impls.py
@@ -151,3 +151,5 @@ class QubitMethods(_interp.MethodTable):
                     f"Qubit allocations exceeded {interp.max_qubits}.",
                 ),
             )
+
+        return (interp.lattice.bottom(),)


src/bloqade/gemini/analysis/logical_validation/impls.py
```diff
@@ -122,3 +121,33 @@ def controlled_gate(
...
+            interp.add_validation_error(
+                stmt,
+                ir.ValidationError(
+                    stmt,
+                    "Cannot determine qubit address location.",
+                ),
+            )
+            return (interp.lattice.bottom(),)
+
+        if qubit_val.data > interp.max_qubits:
```
david-pl: I think there's an off-by-one error here: the largest allowed address for 10 qubits should be `9`, but this compares to a hard-coded `10`, so it will allow up to `11` qubits.


```suggestion
        if qubit_val.data >= interp.max_qubits:
```

test/gemini/analysis/test_logical_validation.py
```diff
@@ -215,3 +212,13 @@ def main():
...
             squin.broadcast.ry(0.123, [q[0], q[0]])
             squin.cx(q[1], q[1])
+
+
+def test_allocation():
+    with pytest.raises(ValidationErrorGroup):
+
+        @gemini.logical.kernel(aggressive_unroll=True)
+        def main():
+            reg = squin.qalloc(12)
```
david-pl: Now try changing that to `squin.qalloc(11)` 😄

PR #736, opened 2026-03-26
add Stim REPEAT support for scf.For loops

johnzl-777:
The second part to #723 , finally accomplishes the desired REPEAT support behavior that has been so elusive.

Some rough edges I'm not too happy with:
- Note that https://github.com/QuEraComputing/kirin/issues/577 causes problems when you want to express more complicated scf.For patterns like the color code memory experiment. 
- Adding measurements to a list that was made via `squin.measure` is fine, but declaring an empty list (even with explicit typing) falls apart. That's why you might see this weird `measure(qalloc(0))` pattern.
- ~~I had to override the default constprop behavior which gives a huge performance boost but the way that override occurs does not seem idiomatic to me.~~
    - I think I figured out a more idiomatic way, but still feels funky to me


src/bloqade/analysis/measure_id/impls.py
```diff
@@ -270,6 +289,39 @@ def if_else(
...
+                loop_vars = interp_.frame_call_region(
+                    body_frame, stmt, stmt.body, NotMeasureId(), *loop_vars
+                )
+
+            for ssa, val in body_frame.entries.items():
+                body_values[ssa] = body_values.setdefault(ssa, val).join(val)
+
+            if loop_vars is None:
+                loop_vars = ()
+            elif isinstance(loop_vars, interp.ReturnValue):
```
david-pl: Is that an early `return` from a loop? I think that's unsupported by kirin.

src/bloqade/stim/analysis/from_squin_validation/analysis.py
```diff
@@ -89,6 +89,37 @@ def if_else(
 @func.dialect.register(key="stim.validate.from_squin")
 class _FuncMethods(interp.MethodTable):
 
+    @interp.impl(func.Invoke)
```
david-pl: Aren't we inlining everything at some point? If so, then I think it would be simpler to validate after the inline.

src/bloqade/stim/dialects/stim_cf/stmts.py
```diff
@@ -0,0 +1,22 @@
...
+from kirin.decl import info, statement
+
+from ._dialect import dialect
+
+
+@statement(dialect=dialect, init=False)
+class Repeat(ir.Statement):
+    name = "repeat"
+    traits = frozenset({ir.HasCFG(), ir.SSACFG()})
+    count: ir.SSAValue = info.argument(type=types.Int)
```
david-pl: I think it might make sense to make this an `attribute`. AFAIK, the STIM REPEAT statement only supports compile time integers, but I'm not sure.

src/bloqade/stim/dialects/stim_cf/stmts.py
```diff
@@ -0,0 +1,22 @@
...
+    body: ir.Region = info.region(multi=False)
+
+    def __init__(self, count: ir.SSAValue, body: ir.Region | None = None):
```
david-pl: Why not just `body: ir.Region = ir.Region(ir.Block())` ?

src/bloqade/stim/passes/cleanup_non_stim.py
```diff
@@ -0,0 +1,36 @@
...
+
+        # Don't remove statements with regions (like scf.For)
+        if node.regions:
+            return RewriteResult()
+
+        # Don't remove terminators (Yield, Return, Branch, etc.)
+        if node.has_trait(ir.IsTerminator):
+            return RewriteResult()
+
+        node.delete()
```
david-pl, line 60: Not a huge fan of this to be honest. This is like DCE that removes impure statements. That means, if a rewrite fails somewhere in the middle, it would delete things like `squin.x`. The output would then be a silently malformed program. This could lead to nasty debugging and judging from past experience it will 😅 

Is there a safer alternative? For example, instead of defining what to keep, define what needs to go (such as the specific statements you mention in the comment above `qubit.New, qubit.Measure, ...`. If you keep adding stuff until all tests here pass it should be fine, right? Even if we miss something now, it would be infinitely easier to just add a statement here if it's missing than tracking down an accidentally deleted one.

david-pl: Another suggestion: if you decide to go for the delete list rather than the keep list, how about you keep the `self.keep: DialectGroup` and issue a warning whenever a statement that is from an unexpected dialect survived?

src/bloqade/stim/passes/constprop_override.py
```diff
@@ -0,0 +1,101 @@
...
+            if prev_loop_vars is not None and loop_vars == prev_loop_vars:
+                break
+            prev_loop_vars = loop_vars
+
+        if not frame_is_not_pure:
+            frame.should_be_pure.add(stmt)
+        return loop_vars
+
+
+def _patch_constprop(propagate: const.Propagate) -> None:
```
david-pl: How much of a hack is this? 😄 

I think a cleaner solution would be to define your own `ConstProp` analysis that has `keys = ("stim.constprop", "constprop")` and register the method above under `"stim.constprop"`. Then, kirin should handle picking the right method.

I'm not sure, however, whether you need this to work with the standard `HintConst` pass. You define your own pass below, which could just use the custom `ConstProp` analysis, but getting it to work with `HintConst` directly might be tricky / hacky.

src/bloqade/stim/passes/flatten.py
```diff
@@ -1,26 +1,108 @@
...
-        rewrite_result = RewriteResult()
-        rewrite_result = self.simplify_if(mt).join(rewrite_result)
-        rewrite_result = self.unroll(mt).join(rewrite_result)
-        return rewrite_result
+        result = RewriteResult()
+
+        # --- simplify ifs ---
+        result = self.simplify_if(mt).join(result)
+
+        # --- fold constants (needed so get_repeat_range sees const range) ---
```
david-pl, line 72: Would this simplify if you made the range a compile time attribute?

src/bloqade/stim/passes/flatten.py
```diff
@@ -1,26 +1,108 @@
...
+
+        # --- ilist canonicalization ---
+        result = (
+            Walk(Chain(ilist.rewrite.ConstList2IList(), ilist.rewrite.Unroll()))
+            .rewrite(mt.code)
+            .join(result)
+        )
+
+        # --- inline function calls ---
+        def inline_heuristic(node: ir.Statement) -> bool:
```
david-pl: Does this mean you are not inlining function calls inside loops (including stdlib calls)? Why?

src/bloqade/stim/passes/hint_const_in_loops.py
```diff
@@ -0,0 +1,215 @@
+"""Set const hints on statements inside preserved scf.For bodies.
+
+HintConst only exposes the top-level const propagation frame, so SSA values
```
david-pl: Could this be combined with the custom pass you have for squin? Would that make sense? If it results in a single huge class keeping them separate may be preferable.

src/bloqade/stim/passes/hint_const_in_loops.py
```diff
@@ -0,0 +1,215 @@
...
+            h = arg.hints.get("const")
+            if not isinstance(h, const.Value):
+                return RewriteResult()
+            arg_vals.append(h.data)
+        node.result.hints["const"] = const.Value(IList(arg_vals))
+        return RewriteResult(has_done_something=True)
+
+
+class HintLen(RewriteRule):
+    """Hint py.Len results as const when the collection has IList type with Literal length."""
```
david-pl: Is kirin properly defining that though? I think a lot of places just have length `Any` even if it could be known.

src/bloqade/stim/rewrite/qubit_to_stim.py
```diff
@@ -1,19 +1,21 @@
...
 from bloqade.stim.dialects import gate as stim_gate, collapse as stim_collapse
-from bloqade.stim.rewrite.util import (
-    insert_qubit_idx_from_address,
-)
+from bloqade.analysis.address import Address
+from bloqade.stim.rewrite.util import insert_qubit_idx_from_address
 
 
+@dataclass
 class SquinQubitToStim(RewriteRule):
```
david-pl: From the naming, I wouldn't expect this to rewrite gate statements.

src/bloqade/stim/passes/cleanup_non_stim.py
david-pl: Ideally, these statements would actually be deleted by different rewrite rules in the pass. If a statement is left at the end, it kind of feels like a rewrite failed. I'm not sure, however, if that's feasible with safe deletion.

src/bloqade/stim/rewrite/scf_for_to_repeat.py
```diff
@@ -0,0 +1,61 @@
...
+        # Replace scf.For results with corresponding initializers.
+        for result, init in zip(node.results, node.initializers):
+            result.replace_by(init)
+
+        # Create count constant
+        count_stmt = ConstInt(value=count)
+        count_stmt.insert_before(node)
+
+        # Create Repeat statement
+        repeat = Repeat(count=count_stmt.result)
```
david-pl: Another point where I'd argue that the `count` should be an attribute: you assert a value is constant and then inject it back into the IR again.

src/bloqade/analysis/measure_id/lattice.py
```diff
@@ -70,6 +70,22 @@ def is_subseteq(self, other: MeasureId) -> bool:
...
+
+    When ConstantFold replaces an IR statement with py.Constant, the original
+    statement's MeasureID handler is lost. This element preserves the constant
+    value so downstream operations (e.g. Add) can interpret it.
+    """
+
+    data: Any
+
+    def is_subseteq(self, other: MeasureId) -> bool:
+        return isinstance(other, ConstantCarrier) and self.data == other.data
```
david-pl, line 88: This is an equality, not `subseteq`. It means that `self.is_subseteq(AnyMeasureId)` will return `False`, but every element must be `subseteq(Top())`.

david-pl (commented):
Overall, looks very good! Just some minor comments and two slightly larger things:
 * I think the `count` on the `Repeat` statement should be an `info.attribute`. That might make a few things simpler.
 * I'm a bit worried about the pass that deletes statements from other dialects left at the end of the pass.

src/bloqade/analysis/measure_id/impls.py
johnzl-777: This pops up in the AddressAnalysis too. I can remove it here in this PR but then I think it's worth propagating the change to other analyses that do this.

src/bloqade/stim/dialects/stim_cf/stmts.py
johnzl-777: No, you're right, it should be the case the integer is immediately available.

src/bloqade/stim/analysis/from_squin_validation/analysis.py
johnzl-777: I thought the same thing in the beginning but then thought this would be nicer in terms of UX because it keeps the program in the form the user originally wrote it (might make things easier to debug). 

No objections to going back to the original approach if that's more desirable

src/bloqade/stim/rewrite/qubit_to_stim.py
johnzl-777: It does seem silly now that I realize most of the `case` statements are all targeting `squin.gate` dialect statements as opposed to `squin.qubit`. More than happy to split off that awkward Reset into its own rule (:

src/bloqade/stim/passes/flatten.py
johnzl-777: No, the stdlib calls are definitely inlined. This is an odd predicate from Claude.

src/bloqade/stim/passes/constprop_override.py
johnzl-777: I agree this is definitely awkward but this is also the least invasive approach I could think of (it was even worse earlier when Claude has this weird "install" method floating around). 

Just to be clear, when you say "define my own constprop" you mean subclass the Kirin one, right? In which case, it would be acceptable to ignore the `@final` it's been decorated with?

src/bloqade/stim/passes/hint_const_in_loops.py
johnzl-777: This does work/do something 👍

src/bloqade/stim/passes/cleanup_non_stim.py
johnzl-777: I agree this rewrite rule is definitely iffy. I opted for your suggestions here to make things safer and tried to delegate more to DCE as well. 

So, to correct your earlier statement now it _is_ DCE that removes impure statements lol.

I actually realized that this just supercedes the functionality of the dead register removal rule I had floating around so I eliminated that too.

johnzl-777: I guess if we really wanted to respect that dead impure statements not be touched/manipulated we could push things into the emission side of things but that seems pretty ugly too.

src/bloqade/stim/passes/flatten.py
johnzl-777: On scf.For?

david-pl: No, sorry, I got confused and was thinking you'd already have `Repeat` here. But that's not the ase.

src/bloqade/stim/passes/constprop_override.py
david-pl: > Just to be clear, when you say "define my own constprop" you mean subclass the Kirin one, right?

No, I mean define a new analysis (inheriting from `Forward`) and just use `keys = ("stim.constprop", "constprop")`. It should then look through the registry for methods under `"stim.constprop"` first and if it doesn't find any fall back to `"constprop"`, which is the behavior you want, right? The only other thing you need is to register the custom method you have here in a new `MethodTable` under the key `"stim.constprop"`.

As I said, the only thing that won't work then is that code that uses the default `HintConst` pass won't have the new method. But you are already defining a custom `StimHintConst` in which you can use it. I'm just not sure whether there is any pass where you rely on the default `HintConst` to have this method.

src/bloqade/stim/analysis/from_squin_validation/analysis.py
david-pl: Well, what's the intent of this validation? If it's only to be used in the context of rewriting to STIM, then why not inline first since that is what happens anyway? You'd then just need to validate that there are no function calls left. Anyway, I don't have a strong preference here.

src/bloqade/analysis/measure_id/impls.py
david-pl: Oh, I suppose that's a leftover then since we didn't always catch that error on kirin side. However, kirin should just not allow it. E.g., try defining the following kernel

```python
@squin.kernel
def main(n: int):
    q = squin.qalloc(n)
    for i in range(n):
        squin.x(q[i])
        return q

    return squin.broadcast.measure(q)
```

johnzl-777: I see! Let me get rid of it on my end, but worth following up with an issue to propagate to everything else.

johnzl-777: Follow up made in #742

src/bloqade/stim/analysis/from_squin_validation/analysis.py
johnzl-777: Actually I think your approach makes more sense, looking at the code again the redundancy of checking for the existence of the same statement in two different ways doesn't seem worth it to me.

src/bloqade/stim/passes/constprop_override.py
johnzl-777: I don't think the impact on default HintConst is a problem here (: but I have gone ahead with your approach for now and after talking with Roger will push for constprop on the Kirin side to embody this optimization

src/bloqade/stim/passes/hint_const_in_loops.py
johnzl-777: I don't think it makes sense to join them considering one specifically handles the loop bodies while the other doesn't

src/bloqade/analysis/measure_id/lattice.py
david-pl: This equality is still a little weird. Could someone draw this lattice for me? 😄 At which point does the `ConstantCarrier` fit into the hierarchy?

I also noticed now that `ConcreteMeasureId` has a similar issue in that it implements `is_subseteq` as equality, meaning it's not `subseteq` to `AnyMeasureId`. But that's a pre-existing bug.

@weinbe58 I think you originally implemented this lattice, correct? What's the reasoning behind this?

src/bloqade/stim/passes/constprop_override.py
```diff
@@ -0,0 +1,177 @@
...
+                break
+            prev_loop_vars = loop_vars
+
+        if not frame_is_not_pure:
+            frame.should_be_pure.add(stmt)
+        return loop_vars
+
+
+@dataclass
+class StimConstProp(ForwardExtra[const.Frame, const.Result]):
```
david-pl, line 76: Can you inherit from `HintConst` here? Or does that mess things up with the registry? Might be able to reduce code duplication with that.

src/bloqade/stim/rewrite/scf_for_to_repeat.py
```diff
@@ -0,0 +1,56 @@
...
+        body_block = node.body.blocks[0]
+
+        stmt = body_block.first_stmt
+        while stmt is not None:
+            next_stmt = stmt.next_stmt
+            if isinstance(stmt, Yield):
+                stmt.delete()
+            else:
+                stmt.detach()
+                repeat_block.stmts.append(stmt)
```
david-pl, line 51: Do you need `stmt.attach(repeat_block)` here?

src/bloqade/stim/passes/repeat_eligible.py
```diff
@@ -0,0 +1,33 @@
...
+    if isinstance(data, range):
+        r = data
+    elif hasattr(data, "data") and isinstance(data.data, range):
+        # IList(range(...)) case
+        r = data.data
+    else:
+        return None
+
+    body_block = node.body.blocks[0]
+    loop_var = body_block.args[0]
```
david-pl, line 29: You only check the first block argument here. This ignores other values passed into the loop. Claude generated a failing test:

```python
def test_set_observable_on_accumulated_iter_arg():
    """Accumulator iter_arg passed to set_observable after the loop.

    The loop body grows `acc` each iteration, so the scf.For result is
    NOT loop-invariant. ScfForToRepeat unconditionally replaces the result
    with the initializer, so set_observable sees only the 2 initial
    measurements instead of all 8.
    """
    @squin.kernel
    def test():
        qs = squin.qalloc(2)
        acc = squin.broadcast.measure(qs)        # 2 initial measurements
        for _ in range(3):
            ms = squin.broadcast.measure(qs)
            acc = acc + ms                        # iter_arg modified
        squin.set_observable(acc)                 # should include all 8

    SquinToStimPass(dialects=test.dialects)(test)
    result = codegen(test)
    # Expected: OBSERVABLE_INCLUDE(0) rec[-8] rec[-7] rec[-6] ... rec[-1]
    # Actual (buggy): OBSERVABLE_INCLUDE(0) rec[-8] rec[-7]
    assert "rec[-1]" in result, f"Accumulator truncated to init:\n{result}"
    for i in range(1, 9):
        assert f"rec[-{i}]" in result, f"Missing rec[-{i}]:\n{result}"
```

david-pl (changes requested):
Looks good overall, just some minor stuff and one bug that needs fixing (but should be easy).

Other than that, it may makes sense to add a few tests for loops that shouldn't be converted (as far as I can tell there aren't any so far).

Good work! 🎉

src/bloqade/stim/rewrite/scf_for_to_repeat.py
johnzl-777: I don't think so, stmt.attach gets invoked via .append and from other parts of the Kirin source it looks like it's the more idiomatic approach

src/bloqade/stim/passes/constprop_override.py
johnzl-777: I'm getting a bit confused here, do you mean subclass HintConst for the `StimHintConst`? I don't think that gains us anything?

src/bloqade/stim/passes/repeat_eligible.py
johnzl-777: Good catch, this should definitely be REPEAT-able. Some more digging with Claude shows you've caught a case I should have seen earlier. Fix doesn't look too bad but requires a bit more thought on how the partial rewrites are being done right now

src/bloqade/stim/passes/constprop_override.py
david-pl: Yes. I was just thinking it might save implementing methods such as `__post_init__`, which should be largely the same. Or am I mistaken? In any case, it's just a small optimization so feel free to skip.

src/bloqade/analysis/measure_id/impls.py
```diff
@@ -202,14 +215,20 @@ def add(self, interp: MeasurementIDAnalysis, frame: MeasureIDFrame, stmt: py.Add
         lhs = frame.get(stmt.lhs)
         rhs = frame.get(stmt.rhs)
 
+        # Unwrap constant carriers holding empty ILists into empty MeasureIdTuples
+        if isinstance(lhs, ConstantCarrier) and isinstance(lhs.data, ilist.IList):
+            lhs = MeasureIdTuple(data=(), obj_type=ilist.IList)
+        if isinstance(rhs, ConstantCarrier) and isinstance(rhs.data, ilist.IList):
+            rhs = MeasureIdTuple(data=(), obj_type=ilist.IList)
```
weinbe58:
```suggestion
        if isinstance(lhs, ConstantCarrier) and isinstance(lhs.data, ilist.IList) and len(lhs.data) == 0:
            lhs = MeasureIdTuple(data=(), obj_type=ilist.IList)
        if isinstance(rhs, ConstantCarrier) and isinstance(rhs.data, ilist.IList) and len(rhs.data) == 0:
            rhs = MeasureIdTuple(data=(), obj_type=ilist.IList)
```

This probably can't happen but its worth double checking here

src/bloqade/analysis/measure_id/lattice.py
weinbe58: ConstCarrier is a simple element it is just a value for `py.Constant`. You need this because you need to inspect statements like `measurement = measurements[-1]` inside the loop body where `measurements` is not a constant value and Constant propagation didn't actually add values inside the body of a for loop. This might be different in the newest patch that @johnzl-777 added.

src/bloqade/stim/analysis/from_squin_validation/analysis.py
```diff
@@ -3,18 +3,56 @@
...
+        ):
+            return False
+        start = owner.start.owner.value.unwrap()
+        stop = owner.stop.owner.value.unwrap()
+        step = owner.step.owner.value.unwrap()
+        return start == 0 and stop >= 1 and step == 1
+    return False
+
+
 class _StimIfElseValidationAnalysis(Forward[EmptyLattice]):
```
david-pl: Nit pick: this class name is no longer accurate.

david-pl (commented):
@johnzl-777 looks good now, I think you addressed all my concerns. Codex found some small bugs:

1. `ResolveAnnotate` has its own observable counter. This can lead to duplicate observable emits.

Here's an MRE

```python
def test_mixed_observable_paths_use_distinct_observable_ids():
    @squin.kernel
    def test():
        qs = squin.qalloc(2)
        m0 = squin.broadcast.measure(qs)
        squin.set_observable([m0[0]])  # SetObservablePartial path

        acc = []
        for _ in range(2):
            ms = squin.broadcast.measure(qs)
            acc = acc + ms

        squin.set_observable(acc)  # ResolveSetAnnotate path

    SquinToStimPass(dialects=test.dialects)(test)
    result = codegen(test)

    assert "OBSERVABLE_INCLUDE(0) rec[-2]" in result
    assert "OBSERVABLE_INCLUDE(1) rec[-4] rec[-3] rec[-2] rec[-1]" in result
```

2. REPEAT eligibility only checks unused loop var + const range, but loop-carried aggregates consumed inside the preserved loop can crash. Repro: inside for _ in range(3), build acc = acc + ms and call squin.set_detector(acc, ...); it raises AttributeError: 'BottomType' object has no attribute 'vars' at src/bloqade/stim/rewrite/set_detector_partial.py:70. set_observable(acc) has the same issue via set_observable_partial.py:36


```python
def test_loop_carried_accumulator_consumed_inside_repeat_loop():
    @squin.kernel
    def test():
        qs = squin.qalloc(2)
        acc = []
        for _ in range(3):
            ms = squin.broadcast.measure(qs)
            acc = acc + ms
            squin.set_detector(acc, coordinates=[0, 0])

    SquinToStimPass(dialects=test.dialects)(test)
    result = codegen(test)

    assert "REPEAT 3" in result
    assert "DETECTOR(0, 0)" in result
```

johnzl-777:
@david-pl I was able to fix both problems, for 1. I went with the most idiomatic approach of a separate analysis (unfortunately this means there's a small amount of code duplication that happens here) and for 2. it seems like there's some more logic for the type hinting to be correct. 

It's kind of funny seeing the Claude-generated accumulation pattern inference code because this is similar to my O(1) but more complicated REPEAT-eligibility analysis 😅

src/bloqade/analysis/observable_idx/analysis.py
```diff
@@ -0,0 +1,44 @@
...
+    """Assigns lexical-order observable indices to SetObservable statements.
+
+    Lightweight Forward analysis whose only job is to number SetObservable
+    statements consistently across rewrite stages. Runs cheaply on
+    EmptyLattice (no per-SSA tracking, no convergence work beyond the
+    trivial). Both SetObservablePartial and ResolveSetAnnotate read indices
+    from the produced frame to share a single observable-index namespace.
+    """
+
+    keys = ["observable_idx"]
```
david-pl:
```suggestion
    keys = ("observable_idx",)
```

src/bloqade/stim/passes/hint_const_in_loops.py
```diff
@@ -0,0 +1,320 @@
...
+                init.type, (types.BottomType, types.AnyType)
+            ):
+                result.type = init.type
+                has_done_something = True
+                _propagate_type_through_uses(result)
+
+        return RewriteResult(has_done_something=has_done_something)
+
+
+def _propagate_type_through_uses(
```
david-pl, line 154: These private helpers are really tough to read 😅

david-pl (approved):
@johnzl-777 I think this is good to go!

src/bloqade/stim/passes/hint_const_in_loops.py
johnzl-777: I agree, I think the additional logic to handle accumulation semantics is worth revisiting because my gut tells me things could be made neater. For now though I'm convinced this is functionally acceptable

PR #751, opened 2026-04-21
Fix bug where SQRT_Y was actually implemented as SQRT_Y_DAG

rafaelha:
With the standard definition of the SQRT_Y gate in the literature, a CNOT gate
image
is converted into the native gate set as
image

See also [this circuit](https://algassert.com/quirk#circuit={%22cols%22:[[%22H%22,%22H%22],[%22%E2%80%A2%22,1,%22X%22],[1,%22%E2%80%A2%22,1,%22X%22],[%22%E2%80%A2%22,%22X%22],[1,%22Y^-%C2%BD%22],[%22%E2%80%A2%22,%22Z%22],[1,%22Y^%C2%BD%22]]}).

However, in squin.native the following was used (the role of SQRT_Y and SQRT_Y_DAG was swapped):
image

This did not raise an error in unit tests, because the repo consistently swapped the roles of SQRT_Y and SQRT_DAG. So internally, this is really a matter of convention, but it produces bugs when interfacing with external software like Stim and Tsim.

I fixed this across the codebase and added unit tests to verify that the actual unitary matrices are correct.


rafaelha:
@codex please review

weinbe58 (approved):
LGTM, @rafaelha and I talked offline about this.

weinbe58:
@rafaelha @david-pl This PR can't be backported to v0.12.* because the phasedXZ gate was not implemented in that version.

PR #752, opened 2026-04-21
Fix factor of 2 in phased xz

rafaelha:
Documentation was misleading. In all cases the args were treated as turns with the exception of the stdlib implementation in the native dialect.

Separately, the `shift` gate was missing a factor of 2 in the input phase.

Added integration tests based on pyqrack.


rafaelha:
@codex please review

PR #757, opened 2026-04-28
Fix cirq ZZPowGate odd-integer exponent emitting X⊗X instead of Z⊗Z

rafaelha:
## Summary

`visit_ZZPowGate` in `src/bloqade/cirq_utils/lowering.py` emitted `X⊗X` for odd-integer exponents (`ZZ`, `ZZ^3`, ...). This is a silent miscompilation: `ZZ = diag(1, -1, -1, 1) = Z⊗Z` (up to global phase), while `X⊗X` is anti-diagonal — a completely different unitary. Fixed by emitting `Z⊗Z` instead.

## Reproducer

```python
import cirq
from bloqade.cirq_utils import load_circuit

q0, q1 = cirq.LineQubit.range(2)
circuit = cirq.Circuit(cirq.ZZ(q0, q1))  # ZZPowGate(exponent=1)

kernel = load_circuit(circuit)
kernel.print()
# Before: emits squin.gate.x on both qubits  (wrong — that's X⊗X)
# After:  emits squin.gate.z on both qubits  (correct — Z⊗Z = ZZ)
```

Made with CC


rafaelha:
@codex review

david-pl (approved):

test/cirq_utils/test_cirq_to_squin.py
```diff
@@ -478,3 +478,33 @@ def test_cirq_roundtrip_state_vector():
...
+    import contextlib
+
+    q = cirq.LineQubit.range(2)
+    circuit = cirq.Circuit(cirq.ZZ(*q) ** exponent)
+
+    kernel = load_circuit(circuit)
+
+    buf = io.StringIO()
+    with contextlib.redirect_stdout(buf):
+        kernel.print()
```
david-pl: Can we get a test that doesn't rely on printing? 😅

david-pl (changes requested):

PR #759, opened 2026-04-28
Fix STIM Swap emit using wrong map and wrong type annotation

rafaelha:
The `Swap` impl in `EmitStimGateMethods` was annotated as a `ControlledTwoQubitGate` and looked up the gate name in `gate_ctrl_2q_map` instead of `gate_2q_map`. It only emitted correctly because `Swap` had been duplicated into `gate_ctrl_2q_map` as a workaround. Removing that duplicate entry would make the impl raise `KeyError` at emit time.

```python
from bloqade.stim.dialects.gate import stmts as stim_gate_stmts
from bloqade.stim.dialects.gate.emit import EmitStimGateMethods

table = EmitStimGateMethods()
print(stim_gate_stmts.Swap.name in table.gate_2q_map)        # True (where it belongs)
print(stim_gate_stmts.Swap.name in table.gate_ctrl_2q_map)   # before: True (duplicate workaround); after: False
```


rafaelha:
@codex please review

david-pl (approved):

test/stim/dialects/stim/test_stim_circuits.py
```diff
@@ -61,6 +61,15 @@ def test_controlled_two_qubit_gates():
 test_gates()
 
 
+def test_swap_uses_gate_2q_map():
```
david-pl: What is this test?

rafaelha: Removed it altogether

PR #760, opened 2026-04-28
Fix CallGraphPass dropping dialect set on callee copies

rafaelha:
`CallGraphPass.unsafe_run` called `original_mt.similar()` without passing `self.dialects`, so callee copies kept their original dialect group instead of the combined one. Rewrites that depended on the new dialects (e.g. `SquinToNative`'s `GateRule`) silently failed to apply inside callee kernels. The sibling `UpdateDialectsOnCallGraph.unsafe_run` already passes `self.dialects`; this aligns the two.

```python
import inspect
from bloqade.rewrite.passes.callgraph import CallGraphPass, UpdateDialectsOnCallGraph

print("similar(self.dialects)" in inspect.getsource(UpdateDialectsOnCallGraph.unsafe_run))  # True
print("similar(self.dialects)" in inspect.getsource(CallGraphPass.unsafe_run))              # before: False, after: True
```

### DIFF
--- a/test/rewrite/test_callgraph.py
+++ b/test/rewrite/test_callgraph.py
@@ -1,0 +1,37 @@
+from kirin import rewrite
+from kirin.rewrite.abc import RewriteRule, RewriteResult
+from kirin.analysis.callgraph import CallGraph
+
+from bloqade import squin
+from bloqade.squin.gate import stmts as squin_gate_stmts
+from bloqade.rewrite.passes import CallGraphPass
+from bloqade.native.dialects import gate as native_gate
+
+
+class TouchHRule(RewriteRule):
+    def rewrite_Statement(self, node) -> RewriteResult:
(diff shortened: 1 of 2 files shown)


rafaelha:
@codex please review

test/rewrite/test_callgraph.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+class TouchHRule(RewriteRule):
+    def rewrite_Statement(self, node) -> RewriteResult:
+        if isinstance(node, squin_gate_stmts.H):
+            return RewriteResult(has_done_something=True)
+        return RewriteResult()
+
+
+def test_callgraph_pass_propagates_dialects_to_callees():
```
david-pl: It's not clear to me what this test proves. It doesn't check any dialect groups, how is that related to the fix?

PR #761, opened 2026-04-28
Fix address-analysis ilist.ForEach impl returning None

rafaelha:
The `map_` impl in `src/bloqade/analysis/address/impls.py` is registered for both `ilist.Map` and `ilist.ForEach`, but only returned a value in the `Map` branch. The `ForEach` path fell through to an implicit `return None`, violating the kirin impl contract that expects a tuple matching `stmt.results` length (which is `()` for `ForEach`).

```python
from bloqade import squin
from bloqade.analysis import address
from bloqade.analysis.address.impls import IListMethods
from kirin.dialects import ilist

@squin.kernel
def apply_h(qb): squin.h(qb)

@squin.kernel
def main():
    q = squin.qalloc(3)
    ilist.for_each(apply_h, q)

a = address.AddressAnalysis(main.dialects)
frame, _ = a.run(main)
(foreach_stmt,) = (s for s in main.callable_region.walk() if isinstance(s, ilist.ForEach))
print(IListMethods().map_(a, frame, foreach_stmt))   # before: None ; after: ()
```


rafaelha:
@codex please review

src/bloqade/analysis/address/impls.py
```diff
@@ -103,6 +103,8 @@ def map_(
         if isinstance(stmt, ilist.Map):
             return (PartialIList(tuple(results)),)
 
+        return ()
```
david-pl, line 106: I think there's about a hundred of those cases 😅 internally, `None` and `()` are handled identically.

david-pl (approved):

src/bloqade/analysis/address/impls.py
Roger-luo: hopefully this will be fixed after rust refactor...

PR #764, opened 2026-04-29
Add syntactic sugar for squin.two_qubit_pauli_channel

### LINKED ISSUE #341: Syntax sugar for squin's two qubit pauli channel

jasonhan3:
## Summary

Closes #341.

Adds syntax sugar for SQUIN two-qubit Pauli channels so users can pass Cirq-style
Pauli product probability dictionaries instead of manually constructing the full
15-entry probability list.

```python
@squin.kernel
def main():
    q = squin.qalloc(2)
    squin.two_qubit_pauli_channel({"XX": 0.1, "YY": 0.2, "ZZ": 0.15}, q[0], q[1])
```

Omitted Pauli products default to `0.0`.

The existing 15-entry probability list form and broadcast-style control/target lists continue to work.

## Changes

- Add custom Python-call lowering for `squin.noise.stmts.TwoQubitPauliChannel`
- Support dict literals keyed by Pauli products such as `"XX"`, `"YY"`, and `"ZZ"`
- Preserve the existing 15-entry list/IList probability form
- Preserve broadcast controls/targets typed as `IList[Qubit, N]`
- Allow single-qubit controls/targets in the simple public syntax by wrapping them into one-element `IList`s during lowering
- Preserve Python list literals for controls/targets as `IList`s for broadcast use
- Centralize `PAULI_PRODUCT_ORDER` and reuse it from Cirq lowering/emission and fidelity analysis
- Correct the two-qubit Pauli probability operand type to `IList[float, Literal[15]]`

## Verification

```sh
uv run --no-sync pytest test/squin/test_two_qubit_pauli_channel_sugar.py test/stim/passes/test_squin_noise_to_stim.py test/cirq_utils/test_squin_noise_to_cirq.py test/analysis/fidelity/test_fidelity.py
uv run --no-sync ruff check src/bloqade/squin/noise/stmts.py src/bloqade/squin/noise/_interface.py src/bloqade/squin/noise/__init__.py src/bloqade/squin/stdlib/simple/noise.py src/bloqade/cirq_utils/emit/noise.py src/bloqade/cirq_utils/lowering.py src/bloqade/squin/analysis/fidelity/impls.py test/squin/test_two_qubit_pauli_channel_sugar.py
```


src/bloqade/squin/stdlib/simple/noise.py
```diff
@@ -61,26 +62,7 @@ def single_qubit_pauli_channel(px: float, py: float, pz: float, qubit: Qubit) ->
...
-    `{IX, IY, IZ, XI, XX, XY, XZ, YI, YX, YY, YZ, ZI, ZX, ZY, ZZ}`
-
-    weighted with the corresponding list of probabilities.
-
-    **NOTE**: The order of the given probabilities must match the order of the list of Pauli products above!
-    """
-    broadcast.two_qubit_pauli_channel(
-        probabilities, ilist.IList([control]), ilist.IList([target])
-    )
+two_qubit_pauli_channel = noise.two_qubit_pauli_channel
```
david-pl, line 65: Er, what's this? This just aliases the wrapper right? That shouldn't be the case.

david-pl: Oh, I see, you can't support this signature in a kernel, right? Hm, well but this makes `two_qubit_pauli_channel` subtly different from all other channels and gates. Not sure if that's so great...

david-pl (changes requested):
I'm not so sure about this change. It's a little weird because it lowers an `ast.Dict`, when kirin actually doesn't support it.

Something like this errors

```python
@squin.kernel
def main():
    q = squin.qalloc(2)
    probs = {"XX": 0.1, "YY": 0.2, "ZZ": 0.15}
    squin.two_qubit_pauli_channel(probs, q[0], q[1])
```

with `BuildError: Cannot lower Dict node: <ast.Dict object at 0x10ae5a4d0>`

What's worse, however, is this silent footgun:

```python
from bloqade import squin

probs = {"XX": 0.1, "YY": 0.2, "ZZ": 0.15}

@squin.kernel
def main():
    q = squin.qalloc(2)
    squin.two_qubit_pauli_channel(probs, q[0], q[1])
```

This seemingly works fine, but inspecting the IR, you see that it silently inserts the actual `dict` object and uses this as argument

```python
func.func @main() -> !py.NoneType {
  ^0(%main_self):
  │ %0 = py.constant.constant 2 : !py.int
  │ %q = func.invoke qalloc(%0) : !py.IList[!py.Qubit, !Any] maybe_pure=False
  │ %1 = py.constant.constant {'XX': 0.1, 'YY': 0.2, 'ZZ': 0.15} : !Bottom
  │ %2 = py.constant.constant 0 : !py.int
  │ %3 = py.indexing.getitem(%q, %2) : !py.Qubit
  │ %4 = py.ilist.new(values=(%3)){elem_type=!py.Qubit} : !py.IList[!py.Qubit, Literal(1,int)]
  │ %5 = py.constant.constant 1 : !py.int
  │ %6 = py.indexing.getitem(%q, %5) : !py.Qubit
  │ %7 = py.ilist.new(values=(%6)){elem_type=!py.Qubit} : !py.IList[!py.Qubit, Literal(1,int)]
  │      squin.noise.two_qubit_pauli_channel(probabilities=%1, controls=%4, targets=%7)
  │ %8 = func.const.none() : !py.NoneType
  │      func.return %8
```

This will only fail later with spurious rewrite failures or an unhelpful error message.

jasonhan3:
Hi David! Sorry for the delay. I was thinking about this; because dict's are not supported in Kirin, if we wanted to support the cases you mentioned in your comment, I don't think there's a clean way to implement this (without supporting dict's in Kirin).

Thus, I think there are maybe two more "simple" solutions that attempt to sidestep using a dict in kirin:
1. Have a python function that the user calls to convert their python dictionary to a list, and then pass that list into the kernel (and you can call the "two_qubit_pauli_channel" kernel). This would involve defining a simple Python function like
```
PAULI_TO_INDEX = {"IX": 0,
        "IY": 1,
        "IZ": 2,
        "XI": 3,
        "XX": 4,
        "XY": 5,
        "XZ": 6,
        "YI": 7,
        "YX": 8,
        "YY": 9,
        "YZ": 10,
        "ZI": 11,
        "ZX": 12,
        "ZY": 13,
        "ZZ": 14}
def convert_twoq_pauli_dict_to_list(twoq_pauli_dict: dict[str, float]) -> list[float]:
     twoq_pauli_list = [0.0 for _ in range(15)]
     for pauli_str, error_prob in twoq_pauli_dict.items():
            if pauli_str in PAULI_TO_INDEX:
                      pauli_index = PAULI_TO_INDEX[pauli_str]
                      twoq_pauli_list[pauli_index] = error_prob
     return twoq_pauli_list
```


3. Allow the user to pass in not a dictionary, but a list of tuples, like
```
[("XX", 0.1), ("YI", 0.2), ("IZ", 0.3)]
```

I have a draft PR for this: https://github.com/QuEraComputing/bloqade-circuit/pull/840. Because I don't do validation in the kernel, there are some edge cases that don't error but might not cause intuitive behavior. But if we wanted to check this, we'd need some validation pass just for the noise kernels, which I'm not sure if we want.

david-pl:
@jasonhan3 It's a much trickier issue than it seems.

I wonder: could we maybe even modify the statement? Something like

```python
N = TypeVar("N", bound=int)

@statement
class TwoQubitPauliChannel(ir.Statement):
    ...
    paulis: ilist.IList[str, N]
    probs: ilist.IList[float, N]
```

where `paulis` is the list of Pauli strings to apply and the `probs` are the corresponding probabilities. A user then just passes in e.g.

```python
@squin.kernel
def main():
    q = squin.qalloc(2)
    paulis = ["XX", "YY", "ZZ"]
    probs = [0.1, 0.2, 0.15]
    squin.two_qubit_pauli_channel(paulis, probs, q[0], q[1])
```

Verification can be done on the statement itself, or within a validation analysis. Although it might not always be possible to validate (e.g. when the parameters are upstream arguments). But in that case, a runtime error should be fine.

What do you think?


Also, side question @zhenrongliew: do you know whether kirin 2 will support dicts?

jasonhan3:
@david-pl I think this sounds good but it doesn't preserve backwards compatibility with the current statement signature right? I'd probably opt for a standard library function that the user can call like

`squin.two_qubit_pauli_channel_shorthand(paulis, probs, q[0], q[1])`

that internally just calls `squin.two_qubit_pauli_channel`.

(I suppose I also have a question: it's not possible to modify a kernel signature to take in multiple input types, right? As something like

```
@squin.kernel()
def test_multitype(test_input: ilist.IList[float, Any] | ilist.IList[ilist.IList[float, Any], Any]):
    return
```

seems to fail during lowering. So it's not possible to support "syntactic sugar" by changing the function definition as you would in Python?)

Is this worth implementing, or if kirin 2 supports dicts, should we just wait for that?

zhenrongliew:
> Also, side question @zhenrongliew: do you know whether kirin 2 will support dicts?

Yes, I think it will be nice for Kirin rust to support dicts. Supporting an immutable typed map would be straightforward. But a faithful Python dict would need more work.

david-pl:
> that internally just calls squin.two_qubit_pauli_channel

@jasonhan3 how do you plan on doing that? Inserting the missing probabilities inside a kernel might be tricky.

>  it doesn't preserve backwards compatibility with the current statement signature right?

No that would be breaking. But I think the concern can be somewhat limited. It's probably not used by anyone directly, and if we account for it in rewrites etc. the user API might not change (e.g. when running noise passes). Still, breaking this is a valid concern and we should carefully consider alternatives.

> I suppose I also have a question: it's not possible to modify a kernel signature to take in multiple input types, right?

Unit types are not supported. But you can accept different signatures in wrappers by using `typing.overload`. I'm sure there's an example of this somewhere in the codebase that you can work off of.

Overall, I think the best approach here right now is:

* Keep the statement as is so we don't introduce any breaking changes just for sugar.
* Use two signatures in the wrapper. Combine it with custom lowering which rewrites two lists into the 15 expected arguments. If the values of the argument are not known, bail with a lowering error and inform the user they need to use the full signature.

The only part I'm not sure about here: we'd have to use the wrapper directly (as it's in this branch right now) instead of wrapping this in a stdlib function due to the overloaded signature. I don't think it's possible to overload a kernel definition.

If this turns out not to be possible for some reason, we can always fall back to your two distinct functions approach.

Or, we wait for kirin rust (thanks for the response @zhenrongliew).

jasonhan3:
> @jasonhan3 how do you plan on doing that? Inserting the missing probabilities inside a kernel might be tricky.

I currently did this by adding the logic in the kernel itself 😅 Probably not the best solution as it modifies the IR but I'm not sure how else we can do this if we allow the strings to be runtime arguments. The implementation can be found here: https://github.com/QuEraComputing/bloqade-circuit/pull/840/changes#diff-3631d5050a594ad42928e1dbb5fd99eb6a09b76aad391f8664889aba72bc79b9R120-R139

>Overall, I think the best approach here right now is:

> Keep the statement as is so we don't introduce any breaking changes just for sugar.
Use two signatures in the wrapper. Combine it with custom lowering which rewrites two lists into the 15 expected arguments. If the values of the argument are not known, bail with a lowering error and inform the user they need to use the full signature.

I attempted to implement what you said (for the nonbroadcasted and broadcasted versions) in this draft PR: https://github.com/QuEraComputing/bloqade-circuit/pull/843

jasonhan3:
Converting to a draft; this issue might be easier to solve if we have support for dict's in Kirin.

Roger-luo:
yeah I agree we should support an immutable map eventually, we don't have to support a fully python compatible `dict` because we are not reinventing python.

PR #765, opened 2026-04-29
Added Wrapper Function to Infer Kernel Qubit Count

### LINKED ISSUE #331: Utility function: "wrap" a kernel
A utility function that encapsulates some kernel that operates on qubits.

For example, if you have some kernel with a signature
```python
def kernel(*qubits:Qubit, **kwargs) -> None
```
aka acting on some list of qubits as a unitary or other operation, make a utility function that "wraps" that for ease of simulation so that one does not need to write the following code snippet every time

```python
def main():
    qbts = bloqade.squin.qubit.new( ## ) 
    kernel( *qbts, **kwargs )
    bits = bloqade.squin.measure(*qbts)
    return bits

main2 = bloqade.circuit.utils.wrap(kernel,**kwargs)
# main == main2
```
where ##, the number of qubits, is computed by evaluating the required number of input qubits in `kernel` or is set as a variable in the wrapper (`nqubits = ###`).

jasonhan3:
## Summary

Closes https://github.com/QuEraComputing/bloqade-circuit/issues/331.

** NOTE: this ONLY works for signatures that are of the form

def kernel(*qubits:Qubit, **kwargs)

and the qubits CANNOT be supplied as kwargs. Otherwise, this implementation will not work.**

This PR adds `squin.wrap`, a small utility for turning a qubit-operating kernel into a simulation-ready entry point. The generated wrapper allocates qubits, invokes the provided kernel, measures all allocated qubits, and returns the measurement results.

Example:

```python
from bloqade import squin
from bloqade.types import Qubit


@squin.kernel
def bell(q0: Qubit, q1: Qubit):
    squin.h(q0)
    squin.cx(q0, q1)


main = squin.wrap(bell)
```

`squin.wrap` can infer the number of qubits from the wrapped kernel's unbound parameters, or accept an explicit `n_qubits` value when additional kernel parameters are bound by keyword:

```python
@squin.kernel
def ansatz(q0: Qubit, q1: Qubit, theta: float):
    squin.rx(theta, q0)
    squin.cx(q0, q1)


main = squin.wrap(ansatz, 2, theta=0.125)
```

## Changes

- Added `squin.wrap(method, n_qubits=None, **kwargs)`.
- Exported `wrap` from `bloqade.squin`.
- Added validation for non-kernel inputs, unexpected keyword arguments, invalid keyword names, negative qubit counts, and argument-count mismatches.
- Added tests covering qubit-count inference, keyword binding, explicit qubit counts, and validation failures.

## Notes

Kirin lowering does not currently support starred calls like `kernel(*qubits)`, so `squin.wrap` generates a tiny wrapper function with explicit qubit indexing, for example `kernel(qubits[0], qubits[1])`, and then compiles that function through the normal `squin.kernel` path.

## Testing

```bash
uv run --no-sync pytest test/squin/test_wrap.py test/squin/test_stdlib_shorthands.py
uv run --no-sync ruff check src/bloqade/squin/utils.py test/squin/test_wrap.py
```


src/bloqade/squin/utils.py
```diff
@@ -0,0 +1,136 @@
...
+        for name, global_name in zip(kwargs, keyword_globals, strict=True)
+    ]
+    call_args = ", ".join(positional_args + keyword_args)
+
+    call_line = (
+        f"    __wrapped_method__({call_args})"
+        if call_args
+        else "    __wrapped_method__()"
+    )
+    source = (
```
david-pl, line 110: Sorry, but no: we're not writing a wrapper kernel function by composing python strings. This is genuinely not a good idea.

david-pl (changes requested):
Generally, I think the feature idea in the issue is not well thought out.

Right now, the boilerplate you have to write to wrap a `kernel` is

```python
@squin.kernel
def wrapped_kernel():
    q = squin.qalloc(10)
    kernel(q)
    return squin.broadcast.measure(q)
```

This is 4 lines, it doesn't get much simpler than that while keeping things generic. In the above, it's very simple to provide the exact arguments we need for each specific case. If we want to provide a utility function for this it either gets quite complicated or we have to make strong assumptions about the signature of the kernel we want to wrap, limiting the usefulness. The code here seems to be opting for the second approach. This is the better of the two IMO.

To implement this, we'd need to do something like this (pseudo code):

```python
from kirin import ir
from kirin.dialects import func

body_block = ir.Block(
    stmts=[
        func.Invoke((n_qubits,), callee=squin.qalloc),  # allocate qubits
        func.Invoke((qubits,), callee=kernel),  # call the actual kernel
        func.Invoke((qubits,), callee=squin.broadcast.measure),  # measure
        func.Return(measurement_results)
    ]
)
body = ir.Region(blocks=[body_block])
code = func.Function(
    sym_name="wrapped",
    signature=...,
    body=ir.Region()
)
wrapped = ir.Method(
    dialects=kernel.dialects,
    code=code,
)
```

The current implementation is writing the kernel by composing strings, which is not the way to go here.

Also, keep in mind that the suggested implementation here only works for the specific kernel structure takes the list of qubits as the first argument. This is quite a specific assumption to have. Furthermore, the complexity added here is far from trivial. And all that, to save 4 lines of boiler plate. To be honest, I'd vote we simply do not implement this feature at all. We'll just have to live with the 4 lines of extra code.

From the issue, it also seems like the main use case is simulation. We might want to think about how to improve the simulator API to handle arguments more gracefully instead. The fundamental problem is, however, that you can only allocate qubits inside a kernel.

weinbe58:
It should be possible to do this by just looking at the Method signature and then constructing a kernel that invokes the wrapped kernel, it would be the equivilant to the hard coded closure:

```python

def wraps(mt: ir.Method, *args):
    code = mt.code
    if (trait := code.get_trait(HasSignature)) is None:
        raise ValueError("expecting a function that has a signature")
        
    signature = trait.get_signature(code)
    qubit_arg_map, other_arg_map, qubit_only_signature = analyze_signature(signature)
    
    wrapper_body = ir.Region([body_block := ir.Block()))
    # 1.  build block with the qubit arguments    
    # 2. insert logicl to add `py.Constant` for the constant args
    # 3. add invoke statement of `mt` with a mixture of the block arguments and the constant ssa values in the body
    # 4. construct the function statement and new method. 

```

cc: @jasonhan3 @david-pl

jasonhan3:
Sorry for the delay again. I'm not so sure if we wanted to implement this feature (if the additional complexity in implementation warrants the implementation); but I spun up a quick draft of an implementation: https://github.com/QuEraComputing/bloqade-circuit/pull/841 based on @weinbe58 's suggestion to use the kernel's signature. This allows the user to interleave qubit and kernel arguments in their kernel definition.

david-pl:
@jasonhan3 I took a look at the draft PR and it really seems to add significant complexity. I also pointed an agent at it and it found a significant bug in minutes. So I'd argue it's not worth it. Keep in mind that this is all to save the ~4 lines of boilerplate shown above.

@weinbe58 any objections?

jasonhan3:
Marking as a draft for now, as the additional complexity needed to implement this feature seems to outweigh the benefits of having it.

PR #779, opened 2026-05-29
fix: allow tsim.Circuit and stim.Circuit init from program string

### LINKED ISSUE #768: `Import tsim` and `from bloqade import tsim` lead to different behavior
When trying load a circuit using tsim as a direct import and when it is imported via Bloqade I see a different behavior. The last line in the next script crashes:

```python
import tsim
from bloqade import tsim as tsim_bloqade

stim_circuit = """
    R 0
    """
works = tsim.Circuit(stim_circuit)
does_not_work = tsim_bloqade.Circuit(stim_circuit)
```

with the following error:

```
AttributeError                            Traceback (most recent call last)
Cell In[34], line 8
      4 stim_circuit = """
      5     R 0
      6     """
      7 works = tsim.Circuit(stim_circuit)
----> 8 does_not_work = tsim_bloqade.Circuit(stim_circuit)

File [redacted].venv/lib/python3.12/site-packages/bloqade/tsim/circuit.py:33, in Circuit.__init__(self, kernel)
     22 def __init__(self, kernel: ir.Method):
     23     """Initialize tsim.Circuit from a kernel.
     24 
     25     This class inherits from `tsim.Circuit`. For the full API reference of
   (...)     31 
     32     """
---> 33     program_text = _codegen(kernel)
     34     super().__init__(program_text)

File [redacted]/.venv/lib/python3.12/site-packages/bloqade/stim/circuit.py:27, in _codegen(mt)
     25 def _codegen(mt: ir.Method) -> str:
     26     """Compile a kernel to STIM program string."""
---> 27     mt = mt.similar()
     28     SquinToStimPass(mt.dialects)(mt)
     29     buf = io.StringIO()

AttributeError: 'str' object has no attribute 'similar'
```

Here are the dependencies in my pyproject file

```toml
[redacted]
requires-python = "==3.12.*"
dependencies = [
    "bloqade>=0.32.0",
    "bloqade-circuit[tsim]==0.14.1",
    "bloqade-tsim==0.1.4",
    "jupyterlab>=4.5.3",
]
```

rafaelha:
## Summary

Fixes #768.

`bloqade.tsim.Circuit` and `bloqade.stim.Circuit` only accepted an `ir.Method` kernel and unconditionally ran `_codegen`, which calls `kernel.similar()`. Passing a STIM program string — which the bare `tsim`/`stim` packages accept directly — raised:

```
AttributeError: 'str' object has no attribute 'similar'
```

This made `from bloqade import tsim; tsim.Circuit("R 0")` behave differently from `import tsim; tsim.Circuit("R 0")`.

## Change

Both `Circuit.__init__` methods now accept either a kernel or a string:

- An `ir.Method` is compiled to program text via `_codegen` (unchanged behavior).
- A `str` (plus any extra positional/keyword args) is forwarded straight to the underlying `tsim.Circuit` / `stim.Circuit` constructor.

The same bug existed in both `tsim` and `stim` wrappers, so both are fixed.

## Tests

Added tests for string initialization to `test/tsim/test_circuit.py` and `test/stim/test_circuit.py`, verifying the string-built circuit matches the native circuit. Developed test-first (confirmed the new tests reproduce the reported `AttributeError` before the fix). Full `test/tsim` and `test/stim` suites pass.

### DIFF
--- a/src/bloqade/stim/circuit.py
+++ b/src/bloqade/stim/circuit.py
@@ -34,16 +34,22 @@ def _codegen(mt: ir.Method) -> str:
 
 
 class Circuit(_Circuit):
-    def __init__(self, kernel: ir.Method):
-        """Initialize stim.Circuit from a kernel.
+    """A `stim.Circuit` that can be built from a squin kernel or a program string."""
+
+    def __init__(self, kernel: ir.Method | str, *args, **kwargs):
+        """Initialize stim.Circuit from a kernel or a STIM program string.
 
         This class inherits from `stim.Circuit`. For the full API reference of
         the underlying circuit class, see:
         https://github.com/quantumlib/Stim/blob/main/doc/python_api_reference_vDev.md#stim.Circuit
 
         Args:
-            kernel: The kernel to compile into a stim.Circuit.
+            kernel: The kernel to compile into a stim.Circuit, or a STIM program
+                string to pass directly to the underlying circuit.
+            *args: Additional positional arguments forwarded to `stim.Circuit`.
+            **kwargs: Additional keyword arguments forwarded to `stim.Circuit`.
 
         """
-        program_text = _codegen(kernel)
-        super().__init__(program_text)
+        if isinstance(kernel, ir.Method):
+            kernel = _codegen(kernel)
+        super().__init__(kernel, *args, **kwargs)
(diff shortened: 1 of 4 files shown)


david-pl (approved):
Sound change. Should we backport this?

PR #790, opened 2026-06-04
Emit SQUIN measurement-controlled if statements to Cirq

### LINKED ISSUE #408: Emit if-statement from squin kernel as cirq classical control
# Background

# Implementation

## Validation

**Relevant files**:

**Relevant files**:

# Minimal example

```python
from bloqade import squin

@squin.kernel
def main():
    q = squin.qalloc(1)
    squin.h(q[0])
    m = squin.measure(q[0])
    if m == 0:
        squin.x(q[0])

circ = emit_circuit(main)
```

Samfresh-ai:
## Summary
- add a Cirq emitter validation pass for the supported measurement-controlled `if` shape
- preserve measurement keys through SQUIN measurement/getitem emission
- emit the single gate in a valid `if` body as a Cirq classically controlled operation

Closes #408.

## Validation
- `uv run pytest test/cirq_utils/test_clifford_to_cirq.py -q` (`25 passed`, one existing deprecation warning)
- `uv run pytest test/cirq_utils -q` (`315 passed, 1 xfailed`, one existing deprecation warning)
- `uv run ruff check src/bloqade/cirq_utils/emit/base.py src/bloqade/cirq_utils/emit/qubit.py src/bloqade/cirq_utils/emit/validation.py test/cirq_utils/test_clifford_to_cirq.py`
- `uv run ruff format --check src/bloqade/cirq_utils/emit/base.py src/bloqade/cirq_utils/emit/qubit.py src/bloqade/cirq_utils/emit/validation.py test/cirq_utils/test_clifford_to_cirq.py`
- `uv run python -m compileall -q src/bloqade/cirq_utils/emit test/cirq_utils/test_clifford_to_cirq.py`
- `git diff --check`

## Notes
AI assistance was used during implementation. I inspected the SQUIN IR shape, made the code changes, and ran the validation above before submitting.

### DIFF
--- a/src/bloqade/cirq_utils/emit/base.py
+++ b/src/bloqade/cirq_utils/emit/base.py
@@ -6,11 +6,36 @@ import cirq
 from kirin import ir, types, interp
 from kirin.emit import EmitABC, EmitFrame
 from kirin.interp import MethodTable, impl
-from kirin.dialects import py, func, ilist
+from kirin.dialects import py, func, ilist, scf
+from kirin.ir.exception import ValidationErrorGroup
 from typing_extensions import Self
 
 from bloqade.squin import kernel
 from bloqade.rewrite.passes import AggressiveUnroll
+
+from .validation import CirqClassicalControlValidation
+
+
+@dataclass(frozen=True)
+class _MeasurementKeyRef:
+    key: cirq.MeasurementKey
+    bitmask: int | None = None
+
+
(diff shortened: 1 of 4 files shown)


Samfresh-ai:
Closing this for now to get back under the unitaryHACK26 4-open-PR limit. The branch is preserved and this can be reopened later if the active queue clears.

PR #792, opened 2026-06-05
Emit simple squin if statements as Cirq classical controls

### LINKED ISSUE #408: Emit if-statement from squin kernel as cirq classical control

caijintaoduolacmeng-star:
Closes #408

Summary:
- Added Cirq emission support for simple measurement-conditioned squin if statements by lowering them to Cirq classical controls.
- Tracks single-qubit measurement results through emit, supports comparisons to 0/1/False/True, and validates unsupported if shapes before emitting.
- Added tests for controlled emission and invalid multi-gate then bodies.

Verification:
- python -m pytest test/cirq_utils/test_clifford_to_cirq.py -q
- python -m pytest test/cirq_utils -q
- python -m ruff check src/bloqade/cirq_utils/emit/base.py src/bloqade/cirq_utils/emit/qubit.py src/bloqade/cirq_utils/validation.py test/cirq_utils/test_clifford_to_cirq.py

### DIFF
--- a/src/bloqade/cirq_utils/emit/base.py
+++ b/src/bloqade/cirq_utils/emit/base.py
@@ -5,12 +5,15 @@ from dataclasses import field, dataclass
 import cirq
 from kirin import ir, types, interp
 from kirin.emit import EmitABC, EmitFrame
+from kirin.ir.exception import ValidationErrorGroup
 from kirin.interp import MethodTable, impl
-from kirin.dialects import py, func, ilist
+from kirin.dialects import py, scf, func, ilist
 from typing_extensions import Self
 
 from bloqade.squin import kernel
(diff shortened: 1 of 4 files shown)


caijintaoduolacmeng-star:
I am withdrawing this pull request to comply with the unitaryHACK26 limit of no more than four open pull requests at one time. Thank you for your time.

PR #794, opened 2026-06-06
Emit if statements as cirq classical controls

### LINKED ISSUE #408: Emit if-statement from squin kernel as cirq classical control

has9sayed:
Adds support for emitting `if` statements from a squin kernel as `cirq.ClassicallyControlledOperation` in the Cirq backend.

 ```python
from bloqade import squin

@squin.kernel
def main():
    q = squin.qalloc(1)
    squin.h(q[0])
    m = squin.measure(q[0])
    if m == 0:
        squin.x(q[0])

circ = emit_circuit(main)
print(circ)
```
outputs-
```
0: ──────H───M───X(conditions=[(q(0) & 1) == 0])───
             ║   ║
q(0): ═══════@═══^═════════════════════════════════
```

Closes: #408

### CODE REVIEW

**Comment by david-pl** [comment]
Can you remove the binary .pyc cache files?

**Comment by has9sayed** [comment]
@david-pl I removed the pyc files. Could you please review the changes when you have a chance?

File: src/bloqade/cirq_utils/classical_control.py [hunk 1]
```diff
@@ -0,0 +1,180 @@
+from dataclasses import dataclass
+
+import cirq
+from kirin import ir, types
+from kirin.dialects import py, scf, ilist
+
+from bloqade.qubit import stmts as qubit
+from bloqade.squin import gate
+
+_ALLOWED_CMP_VALUES = {0, 1, True, False}
```
david-pl: This will just reduce to `{0, 1}`, so you might as well simplify it.

src/bloqade/cirq_utils/emit/scf.py
```diff
@@ -0,0 +1,42 @@
...
+
+@scf.dialect.register(key="emit.cirq")
+class __EmitCirqScfMethods(MethodTable):
+
+    @impl(scf.IfElse)
+    def if_else(self, emit: EmitCirq, frame: EmitCirqFrame, stmt: scf.IfElse):
+        condition = parse_classical_if_condition(stmt.cond)
+        assert condition is not None
+
+        key = emit.measurement_keys[condition.measure.result]
```
david-pl: Maybe make this defensive, so we don't run into an unexpected `KeyError`?

src/bloqade/cirq_utils/classical_control.py
```diff
@@ -0,0 +1,180 @@
...
+    if not isinstance(term, scf.Yield) or len(term.values) != 0:
+        return None
+
+    body_stmts = then_stmts[:-1]
+    gates = [s for s in body_stmts if isinstance(s, gate.stmts.Gate)]
+    if len(gates) != 1:
+        return None
+
+    allowed_non_gate = (
+        qubit.New,
```
david-pl: Having `qubit.New` here is kinda weird. Is there a test for this? I think this should error. What was the motivation of adding it?

src/bloqade/cirq_utils/classical_control.py
```diff
@@ -0,0 +1,180 @@
...
+        return (m, True) if m is not None else None
+    if isinstance(owner, qubit.IsZero):
+        m = _resolve_bare_measure(owner.measurements)
+        return (m, False) if m is not None else None
+    return None
+
+
+def is_single_qubit_measure(measure: qubit.Measure) -> bool:
+    qubits_type = measure.qubits.type
+    if not hasattr(qubits_type, "vars") or len(qubits_type.vars) < 2:
```
david-pl: Can you check the type here too? I know this is handled at the call site, but still only checking attributes is kind of odd since it would pass for any class that has a `var` attribute with length 2.

src/bloqade/cirq_utils/validation/classical_control.py
```diff
@@ -0,0 +1,118 @@
+from typing import Any
+
+from kirin import ir, interp
+from kirin.lattice import EmptyLattice
+from kirin.analysis import Forward
+from kirin.dialects import scf
+from kirin.validation import ValidationPass
+from kirin.analysis.forward import ForwardFrame
+
+from bloqade.cirq_utils.classical_control import (
```
david-pl, line 10: Importing `cirq_utils` here, means that for any import from validation we are now pulling in `cirq`, which is a heavy dependency. Can you move this validation into cirq, or at least any part that depends on having `cirq` imported? It may also help resolve the circular import error which you fixed by importing the validation in the function call when emitting.

david-pl (changes requested):
@has9sayed thanks for this PR! It's in good shape, just some minor issues. I'll assign the issue to you so it's clear that this is claimed.

src/bloqade/cirq_utils/classical_control.py
has9sayed: I added `qubit.New` defensively, assuming extra non-gate IR in the then-branch might need to pass through, but that would incorrectly allow qalloc inside the branch, which doesn't make sense for a static cirq circuit. Removed it.

src/bloqade/cirq_utils/validation/__pycache__/__init__.cpython-313.pyc
david-pl, line 1: These need to go again.

src/bloqade/cirq_utils/classical_control.py
```diff
@@ -0,0 +1,187 @@
...
+
+import cirq
+from kirin import ir, types
+from kirin.dialects import py, scf, ilist
+
+from bloqade.qubit import stmts as qubit
+from bloqade.squin import gate
+from bloqade.types import QubitType
+
+_ALLOWED_CMP_VALUES = {0, 1}
```
david-pl: Hm, now that it's only two values, why not just use this in-line? It's more readable.

src/bloqade/cirq_utils/classical_control.py
```diff
@@ -0,0 +1,187 @@
...
+        return None
+    owner = ssa.owner
+    if isinstance(owner, (py.GetItem, py.indexing.GetItem)):
+        return _resolve_measure_stmt(owner.obj)
+    if isinstance(owner, qubit.Measure):
+        return owner
+    return None
+
+
+def _resolve_bare_measure(ssa: ir.SSAValue) -> qubit.Measure | None:
```
david-pl: This function is largely the same as the above `_resolve_measure_stmt`. Could you combine them?

david-pl (changes requested):
Almost there! Some more small changes and we're good to go.

david-pl (approved):
All comments have been addressed, this is good to go.

Issue #781, opened 2026-05-30
SWAP gates


jon-wurtz:
Add a native SWAP gate
```python
squin.swap(qubit1: Qubit, qubit2: Qubit)
```

jon-wurtz:
Alternatively, a more general swap network
```python
squin.swap(qubits: list[Qubit], permutation:list[int])
```

jon-wurtz:
This is useful for isomorphism operations on codes, eg from Fig. 2 of https://arxiv.org/pdf/2404.19005

<img width="341" height="327" alt="Image" src="https://github.com/user-attachments/assets/5ca4118b-483c-4472-9b93-e9d9bdf80578" />

slant14:
Hello, @jon-wurtz! May I ask which contribution is welcome to this issue? The single pair swap `squin.swap(qubit1: Qubit, qubit2: Qubit)` or the alternative general version with permutation? 

It was just for clarification. Is it possible for me to take this issue? (If no one is assigned to it already?)

weinbe58:
@slant14 I'm currently working on some tools that would help this feature: https://github.com/QuEraComputing/bloqade-lanes/pull/683

This is experimental work though its not entirely clear when this will be available downstream.

slant14:
> [@slant14](https://github.com/slant14) I'm currently working on some tools that would help this feature: [QuEraComputing/bloqade-lanes#683](https://github.com/QuEraComputing/bloqade-lanes/pull/683)
> 
> This is experimental work though its not entirely clear when this will be available downstream.

I will start reading the issue and some part of the code today. Thanks for sharing. If it is finished I can start working on lowering of swap gate if it is needed as you mentioned before. 

Also if there are some issues  on this feature can be delegated to opensource contribution please tell me. Until then I will take a look at ccz gate from another issue.

Issue #780, opened 2026-05-30
cc..ccz gates


jon-wurtz:
Add a multicontrolled cc..ccz gate
```python
squin.ccz(tgt = qb1, ctrls = [ qb2, qb3, ... ])
```

Bonus points for other multicontrolled gates, or even a generic interface of controlled gates, eg
```python
squin.controlled(gate = squin.T, tgt = [qb1, ...], ctrls = [qb2, ...])
```

david-pl:
These should be stdlib functions. We purposely got rid of the generic `Controlled` statement for IR simplicity.

jon-wurtz:
What would the stdlib function look like? Some reduction of ccz to CZs? But then we are left with pattern matching if there is a native ccz gate in some qec gate set...

david-pl:
> Some reduction of ccz to CZs

Basically, yes. Past experience has shown that extending the statement set is usually not the way to go. In what context would you require pattern matching? I.e., what would you be trying to do that would require reading out ccz gates from the IR?

jon-wurtz:
Cross-referencing #781, certain physical operations correspond to completely different logical operations; for example, physically applying some T gates to an [[8,3,2]] implements a logical CCZ gate. To be able to efficiently represent this CCZ gate (or other operations that don't have a clean representation as the composition of CZ + U3) would be very useful when understanding these codes.

slant14:
@david-pl I want to take this issue and work on it as well. As mentioned in the [paper](https://arxiv.org/pdf/2404.19005) CCZ is non-Clifford. Do you plan to have decomposition or error on stim in this case or other design decision is to be taken here?

david-pl:
I think a similar approach to the SWAP gates would work here: a dedicated statement with a decomposition in the `native` stdlib.
In STIM, the same decomposition can be used with our custom tags, where e.g. `S[T]` gets interpreted as a T-gate by simulators that support it, such as tsim.

Also pinging @weinbe58 here. Does this sound good to you, or did you have something specific for `CCZ` in mind?

PR #797, opened 2026-06-11
SWAP gate implementation

slant14:
Related to issue #781


david-pl:
@slant14 thanks for the submission! I'll review it soon.

slant14:
@david-pl Thanks for running the reviews, I am fixing the docstrings and expanding the test coverage now.

slant14:
@david-pl I added the docstrings but about the path coverage some of the lines are in @kernel and some are defensive gaurds, are there any design decisions taken by your company to mitigate it?

david-pl:
@slant14 Don't worry too much about the CI, so long as the unit tests pass it should be fine.

weinbe58:
@david-pl I think we should make this gate a statement that way we can potentially compile that gate statement to moves at the lower level?

david-pl:
@weinbe58 I suppose this depends on the semantics we want to have for SWAP. It's not specified in the ticket, and I asked @jon-wurtz who said that the stdlib approach here would be fine for now.
My reasoning for going for stdlib first is that in the past the added maintenance cost of new statements wasn't worth it. That said, I'm not strictly against it.

So, my question is: should we make it a statement in this PR? Or, leave it as stdlib function and change it later when we need that?

jon-wurtz:
The use case that motivated me writing the issue are [automorphisms in error correction codes](https://journals.aps.org/prxquantum/abstract/10.1103/vf7v-cpq9), where a permutation of the physical qubits results in some logical operation, or some physical operation results in a permutation of logical qubits. A stdlib swap network composed of three CX gates technically hits the use case and requires no further development; however, that precludes any further analysis (due to degeneracies in how a swap gate and swap network is implemented).

A statement of the form `squin.swap(qubits: list[Qubit], permutation:list[int])` fufills the requirement of (user) interpretability, transparent analysis, and efficient lowering (e.g., either to a circuit swap network or physical swaps). Single swap becomes a stdlib (`sqin.swap([qb1,qb2], [1,0])`) and parallel swap likewise is a stdlib. The design decision is at @david-pl's prerogative, due to maintenance costs.

slant14:
@david-pl According to the comment of @jon-wurtz which design decision you see fits better based on the use case , the ease of maintenance and technical debt? 
If there are any need for fixes I can work on it.

src/bloqade/native/stdlib/broadcast.py
```diff
@@ -303,3 +303,18 @@ def cy(controls: ilist.IList[qubit.Qubit, N], targets: ilist.IList[qubit.Qubit,
...
+
+    Decomposed into three CNOTs: SWAP(a, b) = CX(a, b) CX(b, a) CX(a, b).
+
+    Args:
+        qubits1 (ilist.IList[qubit.Qubit, N]): First qubits of each pair.
+        qubits2 (ilist.IList[qubit.Qubit, N]): Second qubits of each pair.
+    """
+    cx(qubits1, qubits2)
+    cx(qubits2, qubits1)
+    cx(qubits1, qubits2)
```
weinbe58:
```suggestion
    sqrt_y(qubits2)
    cz(qubits1, qubits2)
    sqrt_y(qubits1 + qubits2)
    cz(qubits2, qubits1)
    sqrt_y(qubits1 + qubits2)
    cz(qubits1, qubits2)
    sqrt_y(qubits_2)
```

Could you write this out in terms of sqrt(y) and CZ gates? 

I found a slightly better implementation that uses only sqrt(y) [gates](https://algassert.com/quirk#circuit={%22cols%22:[[%22QFT3%22],[%22inputA3%22,1,1,%22+=A3%22],[1,1,1,1,%22Swap%22,%22Swap%22],[1,1,1,%22%E2%80%A6%22,%22%E2%80%A6%22,%22%E2%80%A6%22],[1,1,1,1,%22Y^-%C2%BD%22],[1,1,1,1,%22Z%22,%22%E2%80%A2%22],[1,1,1,1,%22Y^-%C2%BD%22,%22Y^-%C2%BD%22],[1,1,1,1,%22%E2%80%A2%22,%22Z%22],[1,1,1,1,%22Y^-%C2%BD%22,%22Y^-%C2%BD%22],[1,1,1,1,%22Z%22,%22%E2%80%A2%22],[1,1,1,1,%22Y^-%C2%BD%22]]})

weinbe58:
@slant14 Adding the swap gate to the native statements is going to require work downstream on the compiler pipeline lets open up a follow up issue to track the native swap statement.

david-pl:
@slant14 thank you for your patience here! As @weinbe58 said, we will leave this PR mostly as-is for now. You already did add a statement in the squin dialect, but used a native stdlib approach. This is a good middle-ground for now.

Please consider Phil's suggestion regarding a simplified version, which partially cancels out some gates.

I went through the PR and only have one comment: the statement is missing an impl in the [cirq emit method table](https://github.com/QuEraComputing/bloqade-circuit/blob/988d0404b80f2da1a240d549cd53a3f5ce819a93/src/bloqade/cirq_utils/emit/gate.py#L12) and the cirq -> [squin lowering](https://github.com/QuEraComputing/bloqade-circuit/blob/988d0404b80f2da1a240d549cd53a3f5ce819a93/src/bloqade/cirq_utils/lowering.py#L181) roundtrip would also be good to have. Can you add those?
Other than that, very neat PR! Good work.

slant14:
@david-pl I did the stdlib approach after reading your comment in issue #780. I will take into account what @weinbe58 suggested and also add the missing statement as you suggested as soon as possible starting today. 

Thank you all for you feedback, review and collective review of the matter and the time you put in. 

@weinbe58 If you are willing you can open another issue regarding to the native swap implementation and add your consideration and requirements there and I can work on that one as well since I have already got myself familiar with the context of the code in that area.

weinbe58:
> @weinbe58 If you are willing you can open another issue regarding to the native swap implementation and add your consideration and requirements there and I can work on that one as well since I have already got myself familiar with the context of the code in that area.

This would span across multiple repos and we would need to first work from the bottom up starting in bloqade-lanes before we implement anything here. if you're willing to dig into another code base we can talk about it.

slant14:
> > @weinbe58 If you are willing you can open another issue regarding to the native swap implementation and add your consideration and requirements there and I can work on that one as well since I have already got myself familiar with the context of the code in that area.
> 
> This would span across multiple repos and we would need to first work from the bottom up starting in bloqade-lanes before we implement anything here. if you're willing to dig into another code base we can talk about it.

Sure I would be happy to take a look and be more familiar with other projects as well. (Since it is slightly related to my PHD work and research)

src/bloqade/native/stdlib/broadcast.py
```diff
@@ -303,3 +303,22 @@ def cy(controls: ilist.IList[qubit.Qubit, N], targets: ilist.IList[qubit.Qubit,
     sqrt_x(targets)
     cz(controls, targets)
     sqrt_x_adj(targets)
+
+
+@kernel
+def swap(qubits1: ilist.IList[qubit.Qubit, N], qubits2: ilist.IList[qubit.Qubit, N]):
+    """Apply a SWAP gate on pairs of qubits, exchanging their states.
+
+    Decomposed into three CNOTs: SWAP(a, b) = CX(a, b) CX(b, a) CX(a, b).
```
david-pl: Docstring is outdated.

david-pl (changes requested):
@slant14 Thanks for the work here! Just one outdated docstring and then we're good to go!

**Edit**: oh, and the linter CI. You may want to install [pre-commit hooks](https://pre-commit.com).

david-pl (approved):
Thanks for the work @slant14!

PR #811, opened 2026-06-16
Add RemoveEmptyArgGates pass and associated rules

### LINKED ISSUE #516: Clean up gate and broadcast statements with no qubit inputs.
Our current statements take a list of qubits, which can sometimes be empty. We should have a rule that can go through and remove any statements that do not do anything.

mhmohona:
Adds `RemoveEmptyArgGates`, a `CallGraphPass` that removes squin gates, noise channels, and measurements applied to compile-time-empty qubit ILists. Handles stdlib `func.invoke` calls and direct quantum statements (e.g. after inlining). Preserves used empty measurement results, keeps partial two-qubit ops like `cx([], q)`, and does not touch user-defined wrappers or non-qubit empty lists.

Closes: #516


src/bloqade/rewrite/rules/remove_empty_arg_gates.py
```diff
@@ -0,0 +1,240 @@
...
+from bloqade.squin.noise import stmts as noise_stmts
+
+QuantumOperation = (
+    gate_stmts.Gate,
+    noise_stmts.NoiseChannel,
+    qubit.stmts.Measure,
+    qubit.stmts.Reset,
+)
+QUBIT_ARG_NAMES = frozenset(("qubits", "controls", "targets"))
+BROADCAST_MODULE_PREFIXES = (
```
david-pl: This is a bit tricky: we don't have modules in kirin right now, so you can't really eliminate `invoke` statements, without something hacky like this.
A cleaner solution is to rewrite at statement level. Then, dead code elimination should take care of empty stdlib function calls in the IR (see below).

src/bloqade/rewrite/rules/remove_empty_arg_gates.py
```diff
@@ -0,0 +1,240 @@
...
+
+    return True
+
+
+@dataclass
+class RemoveEmptyArgOps(RewriteRule):
+    """Remove squin gates, noise channels, and measurements on empty qubit lists."""
+
+    def rewrite_Statement(self, node: ir.Statement) -> RewriteResult:
+        if isinstance(node, Invoke):
```
david-pl: This branch here should be redundant with the one below, if you run a `Fold` pass as part of the overall rewrite pass to eliminate empty functions.

src/bloqade/rewrite/rules/remove_empty_arg_gates.py
```diff
@@ -0,0 +1,240 @@
...
+            return self._rewrite_invoke(node)
+
+        if isinstance(node, QuantumOperation):
+            return self._rewrite_quantum_operation(node)
+
+        return RewriteResult()
+
+    def _rewrite_quantum_operation(self, node: ir.Statement) -> RewriteResult:
+        qubit_args = iter_named_qubit_args(node)
```
david-pl: I think an explicit `match` statement would be easier to read here, rather than fetching some arguments depending on string arg names.

src/bloqade/rewrite/rules/remove_empty_arg_gates.py
```diff
@@ -0,0 +1,240 @@
...
+        if not replace_empty_results(node):
+            return RewriteResult()
+
+        node.delete()
+        return RewriteResult(has_done_something=True)
+
+
+@dataclass
+class RemoveEffectlessInvokes(RewriteRule):
+    """Remove zero-arg invokes of wrappers that became no-ops."""
```
david-pl: Oh, you basically wrote your own DCE. Just use `kirin.passes.Fold` instead, which should also succeed in deleting empty functions invokes.

test/rewrite/__pycache__/test_remove_empty_arg_gates.cpython-313-pytest-9.0.3.pyc
david-pl, line 1: Cache file, please remove.

src/bloqade/rewrite/passes/remove_empty_arg_gates.py
```diff
@@ -0,0 +1,34 @@
...
+    """Remove squin statements and stdlib calls that act on empty qubit lists."""
+
+    callgraph_pass: CallGraphPass = field(init=False)
+
+    def __post_init__(self):
+        rule = Fixpoint(
+            Walk(
+                Chain(
+                    RemoveEmptyArgOps(),
+                    RemoveEffectlessInvokes(),
```
david-pl: Any reason you didn't use a consecutive `Fold` pass instead of the custom `RemoveEffectlessInvokes`?

david-pl (changes requested):
@mhmohona Thank you for putting in the effort here! Overall, the PR is in good shape, but I have one (more or less) major comment. But it should actually allow you to simplify the PR.

Right now, you remove `func.Invoke` statements of calls into squin's stdlib. I'd suggest to not do that: the statement level rewrite (removing `squin.gate` and `squin.noise` statements with empty qubit args) in combination with `kirin.passes.Fold` should already work. The `CallGraphPass` will walk the entire call tree and remove the statement. `Fold` should then detect functions with an empty body and remove them from the IR.

Please remove the `func.Invoke` rewrite logic. This should also allow you to simplify the statement rewrite (no more shared helper functions needed). Thanks!

**Edit**: Oh, also make sure to run `isort` so CI passes.

src/bloqade/rewrite/passes/remove_empty_arg_gates.py
mhmohona: Ah my mistake ! I removed `RemoveEffectlessInvokes`. and it should be like now -
```
Inline → RemoveEmptyArgOps → DeadCodeElimination
```
right?

david-pl: Basically, yes. Personally I only tested with the `Fold` pass, not sure if that runs any extra things. But if it works with DCE, it should be fine.

src/bloqade/rewrite/passes/remove_empty_arg_gates.py
```diff
@@ -0,0 +1,39 @@
...
+class RemoveEmptyArgGates(passes.Pass):
+    """Remove squin statements that act on compile-time-empty qubit lists."""
+
+    callgraph_pass: CallGraphPass = field(init=False)
+
+    def __post_init__(self):
+        rule = Fixpoint(
+            Walk(
+                Chain(
+                    Inline(
```
david-pl: What's this for? I don't think we need to inline anything.

src/bloqade/rewrite/passes/remove_empty_arg_gates.py
```diff
@@ -0,0 +1,39 @@
...
+                        lambda code: sum(
+                            1 for s in code.walk() if isinstance(s, QuantumOperation)
+                        )
+                        == 1
+                    ),
+                    RemoveEmptyArgOps(),
+                    DeadCodeElimination(),
```
david-pl: On second thought: I think DCE is not enough. Check kirin's `Fold` pass here: https://github.com/QuEraComputing/kirin/blob/95b478e7a3c22cb3550a2f437bfb531983aa6ab4/src/kirin/passes/fold.py#L21

We may need more of its rules in order for this to remove empty functions, but please test it out and use the smallest possible set.

src/bloqade/rewrite/rules/remove_empty_arg_gates.py
```diff
@@ -0,0 +1,128 @@
...
+            return (qubits,)
+        case _:
+            return None
+
+
+def all_qubit_args_empty(args: tuple[ir.SSAValue, ...]) -> bool:
+    return bool(args) and all(is_empty_ilist_value(arg) for arg in args)
+
+
+def replace_empty_results(node: ir.Statement) -> bool:
```
david-pl: I don't fully understand what this is for. Gate statements usually have no result. Is this for measurement statements only? So long as there are no uses, why do we need to replace the result?

src/bloqade/rewrite/rules/remove_empty_arg_gates.py
```diff
@@ -0,0 +1,128 @@
...
+            result_type = ilist.IListType[MeasurementResultType, types.Literal(0)]
+        else:
+            result_type = result.type
+
+        if not isinstance(result_type, types.Generic) or not result_type.is_subseteq(
+            IListType
+        ):
+            return False
+
+    for result in node.results:
```
david-pl: Why are you doing two separate passes here?

david-pl (changes requested):
@mhmohona Good refactor! We're almost there. I think we can remove the `Inline` though and may need some more rules from the `Fold` pass, but not sure.

src/bloqade/rewrite/passes/remove_empty_arg_gates.py
mhmohona: Ah! Now we want like - `RemoveEmptyArgOps > ConstantFold > InlineGetItem > Call2Invoke > DeadCodeElimination`

src/bloqade/rewrite/rules/remove_empty_arg_gates.py
mhmohona: It was used for `measure` like when the result is still used (e.g. `len(measure([])))` . otherwise we just delete with no replacement.

src/bloqade/rewrite/rules/remove_empty_arg_gates.py
mhmohona: Um, I think it is leftover code earlier. updated now

src/bloqade/rewrite/passes/remove_empty_arg_gates.py
```diff
@@ -0,0 +1,42 @@
...
+        """Remove empty-arg quantum ops across the method call graph."""
+        result = RewriteResult()
+        call_graph = CallGraph(mt)
+        all_methods = set(call_graph.edges.keys()) | {mt}
+        mt_map: dict[ir.Method, ir.Method] = {}
+
+        for original_mt in all_methods:
+            new_mt = original_mt if original_mt is mt else original_mt.similar()
+            rule = Fixpoint(
+                Walk(RemoveEmptyArgOps(call_graph=call_graph, callee=original_mt))
```
david-pl: Just for my understadning: is tracking the callee the reason why you can't directly use a `CallgraphPass` anymore? Do you need to track it for empty measurements?

It's a bit of a pity if this is needed, since without it it would be much simpler (just run the `CallgraphPass` on empty args + `Fold` after and loop them both in a fixpoint). Do you think that would also work here?

david-pl (commented):
@mhmohona Thanks. Btw, if you comment on the related issue, I can assign it to you so you can claim the bounty once the PR is merged.

I'm wondering whether this could be even simpler. In the current version, you're basically duplicating the `CallgraphPass` rather than using it. Can you do something like this:

```python
from kirin.passes import Fold

@dataclass
class RemoveEmptyArgGates(passes.Pass):
    def unsafe_run(self, mt: ir.Method):
        result = self.callgraph_pass(mt)  # the pass you had before that just runs the `RemoveEmptyArgOps` rule
        return Fold(mt.dialects)(mt).join(result)
```

This won't allow you to track the `callee` anymore, but I'm wondering whether it's worth it to just ignore this case. What do you think?

mhmohona:
We still use `CallGraphPass`. The reason we track `callee` is to resolve compile-time empty lists that get passed as parameters into stdlib functions, for example, `broadcast.x([])` becoming a `BlockArgument` inside the stdlib body. The same logic applies to measurements, with one extra wrinkle: we also need to preserve `len(measure([]))`.

Right now we run `CallGraphPass + Fold` together. If we dropped `callee`, it wouldn't just break some edge case, it would break the main stdlib path entirely.

david-pl (approved):
All comments have been addressed. Thanks @mhmohona!

PR #812, opened 2026-06-17
feat(squin): Clifford layer optimizer (fixed-CZ frame search)

lsschmid:
## What this adds

`LayerOptimize` — a squin pass that reduces the number of **single-qubit pulse layers** in a Clifford circuit while keeping the 2-qubit (CZ) layers fixed. The unitary is preserved.

```python
from bloqade.squin.passes import LayerOptimize

LayerOptimize(mt.dialects)(mt)   # rewrites the circuit in place
```

## How it works

1. Normalize the circuit to `{√X, √Y, S, CZ}` + trailing Paulis.
2. Search over per-qubit "frames" (swap √X↔√Y via S-conjugation) to make each layer more uniform.
3. Cancel/merge the diagonal `S` gates that commute through the CZ layers.
4. Batch same-type single-qubit gates on disjoint qubits into one broadcast pulse.

A no-regression guard ensures the result is never worse than the input.

## Results

~18% fewer single-qubit layers on average (up to ~33%) on random Clifford circuits, unitary-preserved. The frame search uses a fast layer-count estimate (no IR per candidate), so it scales.

## Tests

- Unitary equivalence (cirq/stim) on random circuits
- Layer-count benchmark gate (no regression + net improvement)
- Cold-import guard for the `squin.passes` import graph

### DIFF
--- a/src/bloqade/squin/layer_optimizer/__init__.py
+++ b/src/bloqade/squin/layer_optimizer/__init__.py
@@ -1,0 +1,6 @@
+"""Layer-optimization helpers for Clifford circuits.
+
+Pure cirq leaves used by ``LayerOptimize``: ``schedule`` (per-CZ-gap extraction
+and frame materialization), ``simplify`` (combine diagonals through CZ), ``cost``
+(fast layer-count predictor), and ``search`` (bounded-uphill frame search).
+"""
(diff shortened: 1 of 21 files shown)


lsschmid:
Addressed the Copilot review feedback:

- `pierce_into_fixed` precedence check: `assert` → explicit `raise RuntimeError` (no longer skipped under `python -O`).
- `SquinBatchBroadcastsRule._combine_qubits` disjoint-qubit check: `assert` → explicit `raise ValueError` (+ docstring updated).
- `test_import_order`: added a `timeout=120` to `subprocess.run` and included stdout in the failure message.

Tests and linters pass.

lsschmid:
@rafaelha any updates/comments on that?

PR #817, opened 2026-06-18
Add opt-in TICK insertion to Squin → Stim lowering (#816)

### LINKED ISSUE #816: Option to insert TICK between gates in Squin → Stim to preserve execution-order layering in diagrams
## Summary

Add an option to the Squin → Stim transformation that automatically inserts a `TICK` between every Squin gate (i.e. between consecutive gate statements / each gate "layer"), so that the execution order of the original Squin program is preserved in the emitted Stim circuit.

## Background / current behavior

When a Squin program is lowered to Stim via `SquinToStimPass` (`src/bloqade/stim/passes/squin_to_stim.py`), the resulting Stim circuit contains the gate instructions but no `TICK` separators between independent gates (the `TICK` statement already exists — `bloqade.stim.dialects.auxiliary.stmts.annotate.Tick`, emitted as the `TICK` line).

`TICK` is used to denote different layers of gates, otherwise, STIM can reorder the gates when rendering the circuit diagrams. 

## Problem

For circuits where the *authored* layering carries meaning — e.g. logical state-preparation routines, transversal gate schedules, or any circuit where we want the diagram to mirror the physical/logical execution order — the ASAP rendering misaligns the layers. Gates that were written as distinct sequential steps get visually merged, making the diagram harder to read and harder to validate against the intended schedule.

This came up downstream in **bloqade-lanes** (QuEraComputing/bloqade-lanes#641): after broadcasting the Steane state-prep so it runs in parallel across logical qubits, the Stim diagram still renders in ASAP order and the layers don't line up with the actual gate-execution order.

## Proposed solution

Add an opt-in flag (e.g. `insert_ticks: bool = False`) to `SquinToStimPass` (and/or the user-facing `squin_to_stim` entry point) that inserts a `Tick` statement between every Squin gate before/after the gate-lowering rewrites.

With the option enabled, each gate (or each authored gate layer) is separated by a `TICK`, forcing Stim to render one moment per step and aligning the diagram's columns with the program's execution order.

## Implementation sketch

- A rewrite/walk that inserts `auxiliary.stmts.Tick` between consecutive gate statements (after flattening, so ordering is well-defined).
- Gate this behind the new flag so default behavior is unchanged.
- Decide the granularity: a `TICK` after **every** gate statement, vs. one `TICK` between authored "layers" (broadcast gates already represent a parallel layer, so per-statement is likely the natural unit). Worth a short discussion in the issue.

## Considerations / open questions

- **Default off** to avoid changing existing Stim output / detector-and-observable record indexing for current users.
- Interaction with measurements, detectors, observables, and noise channels — inserting `TICK`s must not change measurement record indices or detector semantics (only timing/layout).
- Whether the unit should be "per gate statement" or "per parallel layer" (broadcast op).
- Whether this belongs in the pass itself or as a separate, composable rewrite that callers can run after `SquinToStimPass`.

## Acceptance criteria

- An option exists to enable automatic `TICK` insertion in the Squin → Stim lowering.
- With the option enabled, the emitted Stim circuit has `TICK`s separating gates such that the rendered diagram's columns follow program/execution order rather than ASAP packing.
- Default behavior (option disabled) is unchanged; existing tests pass.
- Detector/observable/measurement record indexing is unaffected by the inserted `TICK`s.

## Context

- Downstream consumer: QuEraComputing/bloqade-lanes#641
- Affected version: bloqade-circuit `0.14.x`

weinbe58:
Closes #816.

## What

Adds an opt-in `insert_ticks` flag to the Squin → Stim lowering that inserts a `TICK` after every gate, reset, measurement, and noise operation, so the emitted Stim circuit preserves the **authored execution-order layering** (one moment per step) instead of being ASAP-packed when rendered as a diagram. Default **off**, so existing output is unchanged.

This came up downstream in bloqade-lanes#641, where broadcasting Steane state-prep across logical qubits still rendered in ASAP order with misaligned layers.

## How

- New standalone, idempotent `InsertTicks` rewrite (`bloqade/stim/rewrite/insert_ticks.py`). It's an atomic per-statement rule driven by `Walk`; an "operation" is any statement in the `stim.gate` / `stim.collapse` / `stim.noise` dialects. Auxiliary statements (constants, records, detectors, observables, coordinates) are skipped, and since constants fold into the operation's text at emit time they never produce stray lines.
- `insert_ticks: bool = False` flag on `SquinToStimPass` (runs the rewrite after the cleanup fixpoint, so DCE leaves the no-op `TICK`s alone) and on the `squin_to_stim` entry point.
- Exposed as a keyword-only arg on both `bloqade.tsim.Circuit` and `bloqade.stim.Circuit` (threaded through shared `_codegen`), so a TICK-layered diagram can be requested when building a circuit straight from a squin kernel. Ignored when building from a program string.
- Drive-by: fixed a copy-paste typo in `stim.Circuit`'s docstring (`*args`/`**kwargs` said "forwarded to `tsim.Circuit`").

## Acceptance criteria

- [x] Opt-in flag enables automatic `TICK` insertion in the Squin → Stim lowering.
- [x] With the flag on, emitted circuit has `TICK`s separating operations so diagram columns follow program order.
- [x] Default behavior (flag off) unchanged; existing tests pass.
- [x] Detector/observable/measurement record indices are unaffected by the inserted `TICK`s (verified: identical `rec[...]` indices with and without the flag).

## Notes / decisions

- **Granularity:** per *operation* (not per authored "layer"). Per @weinbe58, operations that would share a parallel layer can't actually run in parallel here, so per-operation is the right unit.
- **Trailing TICK:** the simple per-statement rule emits a `TICK` after the final operation too. Benign/valid Stim; left as-is for now (suppressing it cleanly is a bit tricky and not worth it yet).

## Tests

- `test/stim/passes/test_insert_ticks.py` — default-unchanged, per-operation separation, record-index preservation, ticks inside `REPEAT`, rewrite idempotency.
- `insert_ticks` cases added to `test/stim/test_circuit.py` and `test/tsim/test_circuit.py`.
- Full `test/stim/` + `test/tsim/` suites: 256 passed, 2 xfailed, no regressions.

### DIFF
--- a/src/bloqade/stim/rewrite/insert_ticks.py
+++ b/src/bloqade/stim/rewrite/insert_ticks.py
@@ -26,9 +26,11 @@ class InsertTicks(RewriteRule):
 
     @staticmethod
     def is_operation(stmt: ir.Statement) -> bool:
+        """Whether the statement is a moment-advancing Stim operation."""
         return getattr(stmt, "dialect", None) in OPERATION_DIALECTS
 
     def rewrite_Statement(self, node: ir.Statement) -> RewriteResult:
+        """Insert a ``TICK`` after ``node`` when it is an un-ticked operation."""
         if not self.is_operation(node):
             return RewriteResult()
 
--- a/src/bloqade/stim/rewrite/insert_ticks.py
+++ b/src/bloqade/stim/rewrite/insert_ticks.py
@@ -1,0 +1,40 @@
+from kirin import ir
+from kirin.rewrite.abc import RewriteRule, RewriteResult
+
+from bloqade.stim.dialects import auxiliary
+
+# Stim dialects whose statements advance a circuit "moment" / time step. A TICK
+# after every such operation forces Stim to render one moment per step instead
+# of ASAP-packing the diagram. Statements in the auxiliary dialect (constants,
+# records, detectors, observables, coordinate annotations) and control flow do
+# not advance a moment, so they are left untouched -- measurement-record and
+# detector indexing are unaffected.
+OPERATION_DIALECTS = frozenset({"stim.gate", "stim.collapse", "stim.noise"})
+
+
+class InsertTicks(RewriteRule):
+    """Insert a ``TICK`` after every Stim operation.
+
+    Operates on a lowered Stim circuit. After each operation statement (gate,
+    reset, measurement, or noise channel) a
+    :class:`~bloqade.stim.dialects.auxiliary.stmts.Tick` is inserted, so the
+    rendered diagram's columns follow program order rather than ASAP packing.
+
+    The rule is idempotent -- an operation already followed by a ``TICK`` is
+    left untouched -- so it is safe under :class:`~kirin.rewrite.Fixpoint`.
+    """
+
+    @staticmethod
+    def is_operation(stmt: ir.Statement) -> bool:
+        dialect = getattr(stmt, "dialect", None)
+        return dialect is not None and dialect.name in OPERATION_DIALECTS
+
+    def rewrite_Statement(self, node: ir.Statement) -> RewriteResult:
+        if not self.is_operation(node):
+            return RewriteResult()
+
+        if isinstance(node.next_stmt, auxiliary.stmts.Tick):
+            return RewriteResult()
+
+        auxiliary.stmts.Tick().insert_after(node)
+        return RewriteResult(has_done_something=True)
--- a/src/bloqade/stim/rewrite/insert_ticks.py
+++ b/src/bloqade/stim/rewrite/insert_ticks.py
@@ -1,7 +1,7 @@
 from kirin import ir
 from kirin.rewrite.abc import RewriteRule, RewriteResult
 
-from bloqade.stim.dialects import auxiliary
+from bloqade.stim.dialects import gate, noise, collapse, auxiliary
 
 # Stim dialects whose statements advance a circuit "moment" / time step. A TICK
 # after every such operation forces Stim to render one moment per step instead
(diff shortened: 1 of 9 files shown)


src/bloqade/stim/rewrite/insert_ticks.py
```diff
@@ -0,0 +1,40 @@
...
+
+from bloqade.stim.dialects import auxiliary
+
+# Stim dialects whose statements advance a circuit "moment" / time step. A TICK
+# after every such operation forces Stim to render one moment per step instead
+# of ASAP-packing the diagram. Statements in the auxiliary dialect (constants,
+# records, detectors, observables, coordinate annotations) and control flow do
+# not advance a moment, so they are left untouched -- measurement-record and
+# detector indexing are unaffected.
+OPERATION_DIALECTS = frozenset({"stim.gate", "stim.collapse", "stim.noise"})
```
weinbe58: using the dialect name is fragile here as the name could be updated. use the object instead, statements can only belong to a single dialect anyways so using the object is fine.

weinbe58: Good call — switched to comparing the `gate`/`collapse`/`noise` dialect objects by identity (frozenset of dialect objects) instead of name strings. Fixed in f33ecf2.

src/bloqade/stim/rewrite/insert_ticks.py
```diff
@@ -0,0 +1,41 @@
+from kirin import ir
+from kirin.rewrite.abc import RewriteRule, RewriteResult
+
+from bloqade.stim.dialects import gate, noise, collapse, auxiliary
+
+# Stim dialects whose statements advance a circuit "moment" / time step. A TICK
```
rafaelha, line 6: maybe this comment is a bit too verbose

rafaelha (approved):

weinbe58:
@david-pl  I tried to backport this PR but it failed with conflicts I am removing the backport tag we will just need to wait for the new version to integrate this change.

PR #820, opened 2026-06-22
Ignore annotations for Cirq and PyQrack backends

### LINKED ISSUE #628: Allow `annotate` statements to work with Cirq/PyQrack
Tracking this from the team discussion concluding #616 , it should be possible to let a kernel with `annotate` statements be executed in Cirq/PyQrack as long as they're ignored with some additional logic for the interpreter.

jackson5689:
Fixes #628.

This adds no-op implementations for decoder annotation statements when lowering/running through the Cirq emit and PyQrack backends. The methods return one placeholder value per statement result so Kirin frame assignment can continue while the backends ignore detector/observable metadata.

Tests:
- `.venv/bin/python -m pytest test/pyqrack/test_target.py test/cirq_utils/test_cirq_to_squin.py::test_squin_annotations_are_ignored_when_emitting_cirq -q`
- `.venv/bin/python -m py_compile src/bloqade/cirq_utils/emit/annotate.py src/bloqade/pyqrack/squin/annotate.py`
- `git diff --check`

Disclosure: this patch was prepared with AI assistance and manually reviewed before submission.

### DIFF
--- a/src/bloqade/cirq_utils/emit/__init__.py
+++ b/src/bloqade/cirq_utils/emit/__init__.py
@@ -5,5 +5,11 @@ Imports dialect emit modules to register method tables and exposes
 """
 
 # NOTE: just to register methods
-from . import scf as scf, gate as gate, noise as noise, qubit as qubit
+from . import (
+    scf as scf,
+    gate as gate,
+    noise as noise,
+    qubit as qubit,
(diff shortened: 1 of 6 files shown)


david-pl (approved):
LGTM, thanks.

PR #832, opened 2026-06-29
Update to pyqrack 2.0

### LINKED ISSUE #770: Breaking changes in `pyqrack` 2
Pyqrack changed majors last week (from 1 to 2). According to its `pyproject.toml` file, the repo is pulling `pyqrack-cpu>=1.70.4` and this leads to breaking changes on the simulator. For example, the following fails:

```python
from bloqade import squin, pyqrack

@squin.kernel
def allocation_example():
    qreg = squin.qalloc(2)

sim = pyqrack.StackMemorySimulator()
task = sim.task(allocation_example)
results = task.run()
```

In the short run, maybe we can update the pyproject file to instead have the following line?
```
"pyqrack-cpu~=1.70"
```
In case it could be helpful I added this change in #771

david-pl:
Closes #770.


src/bloqade/pyqrack/base.py
```diff
@@ -14,44 +14,36 @@
...
-        isTensorNetwork=False,
-        isSchmidtDecomposeMulti=True,
-        isSchmidtDecompose=True,
-        isStabilizerHybrid=False,
-        isBinaryDecisionTree=False,
-        isPaged=True,
-        isCpuGpuHybrid=True,
-        isOpenCL=True,
+        qubit_count=-1,
+        is_schmidt_decompose_multi=True,
```
jasonhan3, line 43: note that this differs from the default in pyqrack: https://pyqrack.readthedocs.io/en/latest/autoapi/pyqrack/index.html#pyqrack.QrackSimulator

jasonhan3 (approved):
Looks good! Small comment on default, and also CI needs to pass

edit: noticed CI was failing on two tests; probably want to change them as follows:

test_cirq_to_squin.py:376: `assert np.allclose(ket[1:3], 0, atol=1e-7)

test_squin2native.py:51: `cirq.testing.assert_allclose_up_to_global_phase(
    old_sv, new_sv, atol=1e-6
)`

test/native/upstream/test_squin2native.py
jasonhan3: probably change to `cirq.testing.assert_allclose_up_to_global_phase(
    old_sv, new_sv, atol=1e-6
)`

src/bloqade/pyqrack/base.py
david-pl: Yes, but the same we had before and I suppose @weinbe58 put some thought into this.

PR #838, opened 2026-07-07
feat: added more descriptive tsim import error msg

### LINKED ISSUE #837: Unclear error message when importing bloqade-tsim==0.1.4
With the new version of bloqade-tsim==0.1.4, which no longer installs "ipywidgets" by default, we can an import error from "pyzx_param". However, the error message is masked by the bloqade.tsim ImportError override.

## To reproduce
1. Run `uv sync --no-dev`
2. Run `uv pip install bloqade-tsim==0.1.4`
3. Run the following code in a notebook
```
from bloqade.tsim import Circuit
from bloqade import squin

@squin.kernel
def test_kernel():
    reg = squin.qalloc(5)
    return

test_circ = Circuit(test_kernel)
```

This will yield the following error

```
ImportError                               Traceback (most recent call last)
Cell In[3], line 1
----> 1 test_circ = Circuit(test_kernel)

File ~/Desktop/qmain/personal-workspace/bloqade-circuit/src/bloqade/tsim/circuit.py:46, in Circuit.__init__(self, kernel, insert_ticks, *args, **kwargs)
     44 if isinstance(kernel, ir.Method):
     45     kernel = _codegen(kernel, insert_ticks=insert_ticks)
---> 46 super().__init__(kernel, *args, **kwargs)

File ~/Desktop/qmain/personal-workspace/bloqade-circuit/src/bloqade/tsim/circuit.py:13, in _MissingTsimCircuit.__init__(self, *args, **kwargs)
     12 def __init__(self, *args, **kwargs):
---> 13     raise ImportError(
     14         "tsim is required for bloqade.tsim.Circuit. "
     15         'Install with: pip install "bloqade-circuit[tsim]"'
     16     )

ImportError: tsim is required for bloqade.tsim.Circuit. Install with: pip install "bloqade-circuit[tsim]"
```

## Issue
The problem is that the user has already installed tsim but the ImportError message says that they should install tsim to fix the problem. The problem is not with installing tsim; the problem can be seen if we import tsim directly. Running the code

```
from tsim import Circuit
```

gives the following

```
ModuleNotFoundError                       Traceback (most recent call last)
Cell In[4], line 1
----> 1 from tsim import Circuit

File ~/Desktop/qmain/personal-workspace/bloqade-circuit/.venv/lib/python3.12/site-packages/tsim/__init__.py:19
      1 """ZX-calculus based circuit sampler for universal quantum circuits.
      2 
      3 tsim supports fast sampling from quantum circuits with few non-Clifford gates.
   (...)     14 - `utils/`: Utility functions for visualization and linear algebra.
     15 """
     17 __version__ = "0.1.4"
---> 19 from tsim.circuit import Circuit as Circuit
     20 from tsim.sampler import (
     21     CompiledDetectorSampler as CompiledDetectorSampler,
     22     CompiledMeasurementSampler as CompiledMeasurementSampler,
     23 )

File ~/Desktop/qmain/personal-workspace/bloqade-circuit/.venv/lib/python3.12/site-packages/tsim/circuit.py:14
      3 from __future__ import annotations
      5 from typing import (
      6     TYPE_CHECKING,
      7     Any,
   (...)     11     overload,
     12 )
---> 14 import pyzx_param as zx
     15 import stim
     16 from pyzx_param.graph.base import BaseGraph

File ~/Desktop/qmain/personal-workspace/bloqade-circuit/.venv/lib/python3.12/site-packages/pyzx_param/__init__.py:48
     46 from . import tikz
     47 from . import simulate
---> 48 from . import editor
     49 from . import routing
     50 from . import local_search

File ~/Desktop/qmain/personal-workspace/bloqade-circuit/.venv/lib/python3.12/site-packages/pyzx_param/editor.py:38
     35 from .editor_actions import MATCHES_VERTICES, MATCHES_EDGES, operations, operations_to_js
     37 if get_mode() == 'notebook':
---> 38 	import ipywidgets as widgets
     39 	from traitlets import Unicode, validate, Bool, Int, Float
     40 	from IPython.display import display, HTML

ModuleNotFoundError: No module named 'ipywidgets'
```

So the true problem is that the user needs to install 'ipywidgets'.

## Proposed solution
Forward the error message that the user needs to install 'ipywidgets' in the bloqade.tsim wrapper.

jasonhan3:
closes https://github.com/QuEraComputing/bloqade-circuit/issues/837

Now, the error message is 

```
ModuleNotFoundError                       Traceback (most recent call last)
File ~/Desktop/qmain/personal-workspace/bloqade-circuit/src/bloqade/tsim/circuit.py:6
      5 try:
----> 6     import tsim
      8     _Circuit = tsim.Circuit

File ~/Desktop/qmain/personal-workspace/bloqade-circuit/.venv/lib/python3.12/site-packages/tsim/__init__.py:19
     17 __version__ = "0.1.4"
---> 19 from tsim.circuit import Circuit as Circuit
     20 from tsim.sampler import (
     21     CompiledDetectorSampler as CompiledDetectorSampler,
     22     CompiledMeasurementSampler as CompiledMeasurementSampler,
     23 )

File ~/Desktop/qmain/personal-workspace/bloqade-circuit/.venv/lib/python3.12/site-packages/tsim/circuit.py:14
      5 from typing import (
      6     TYPE_CHECKING,
      7     Any,
   (...)     11     overload,
     12 )
---> 14 import pyzx_param as zx
     15 import stim

File ~/Desktop/qmain/personal-workspace/bloqade-circuit/.venv/lib/python3.12/site-packages/pyzx_param/__init__.py:48
     47 from . import simulate
---> 48 from . import editor
     49 from . import routing

File ~/Desktop/qmain/personal-workspace/bloqade-circuit/.venv/lib/python3.12/site-packages/pyzx_param/editor.py:38
     37 if get_mode() == 'notebook':
---> 38 	import ipywidgets as widgets
     39 	from traitlets import Unicode, validate, Bool, Int, Float

ModuleNotFoundError: No module named 'ipywidgets'

The above exception was the direct cause of the following exception:

ImportError                               Traceback (most recent call last)
Cell In[3], line 1
----> 1 test_circ = Circuit(test_kernel)

File ~/Desktop/qmain/personal-workspace/bloqade-circuit/src/bloqade/tsim/circuit.py:48, in Circuit.__init__(self, kernel, insert_ticks, *args, **kwargs)
     46 if isinstance(kernel, ir.Method):
     47     kernel = _codegen(kernel, insert_ticks=insert_ticks)
---> 48 super().__init__(kernel, *args, **kwargs)

File ~/Desktop/qmain/personal-workspace/bloqade-circuit/src/bloqade/tsim/circuit.py:14, in _MissingTsimCircuit.__init__(self, *args, **kwargs)
     13 def __init__(self, *args, **kwargs):
---> 14     raise ImportError(
     15         "tsim is required for bloqade.tsim.Circuit. "
     16         'Install with: pip install "bloqade-circuit[tsim]". '
     17         f"Tsim import error: {_TSIM_IMPORT_ERROR}"
     18     ) from _TSIM_IMPORT_ERROR

ImportError: tsim is required for bloqade.tsim.Circuit. Install with: pip install "bloqade-circuit[tsim]". Tsim import error: No module named 'ipywidgets'
```

### DIFF
--- a/src/bloqade/tsim/circuit.py
+++ b/src/bloqade/tsim/circuit.py
@@ -7,15 +7,16 @@ try:
 
     _Circuit = tsim.Circuit
 except ImportError as _tsim_import_error:
-    _TSIM_IMPORT_ERROR = _tsim_import_error
 
     class _MissingTsimCircuit:
+        _import_error = _tsim_import_error
+
         def __init__(self, *args, **kwargs):
             raise ImportError(
                 "tsim is required for bloqade.tsim.Circuit. "
                 'Install with: pip install "bloqade-circuit[tsim]". '
-                f"Tsim import error: {_TSIM_IMPORT_ERROR}"
-            ) from _TSIM_IMPORT_ERROR
+                f"Tsim import error: {self._import_error}"
+            ) from self._import_error
 
     _Circuit = _MissingTsimCircuit
 
--- a/src/bloqade/tsim/circuit.py
+++ b/src/bloqade/tsim/circuit.py
@@ -6,14 +6,16 @@ try:
     import tsim
 
     _Circuit = tsim.Circuit
-except ImportError:
+except ImportError as _tsim_import_error:
+    _TSIM_IMPORT_ERROR = _tsim_import_error
 
     class _MissingTsimCircuit:
         def __init__(self, *args, **kwargs):
             raise ImportError(
                 "tsim is required for bloqade.tsim.Circuit. "
-                'Install with: pip install "bloqade-circuit[tsim]"'
-            )
+                'Install with: pip install "bloqade-circuit[tsim]". '
+                f"Tsim import error: {_TSIM_IMPORT_ERROR}"
+            ) from _TSIM_IMPORT_ERROR
 
     _Circuit = _MissingTsimCircuit
 


src/bloqade/tsim/circuit.py
```diff
@@ -6,14 +6,16 @@
...
 
     class _MissingTsimCircuit:
         def __init__(self, *args, **kwargs):
             raise ImportError(
                 "tsim is required for bloqade.tsim.Circuit. "
-                'Install with: pip install "bloqade-circuit[tsim]"'
-            )
+                'Install with: pip install "bloqade-circuit[tsim]". '
+                f"Tsim import error: {_TSIM_IMPORT_ERROR}"
+            ) from _TSIM_IMPORT_ERROR
```
david-pl: Isn't `raise from` and printing the error inside the error message a bit redundant? Also, why do we need to bind `_TSIM_IMPORT_ERROR = _tsim_import_error`? Couldn't we just use `_tsim_import_error` directly?

jasonhan3: Yeah, I agree that it is rendundant, but the problem is that the import error isn't always from not having installed bloqade-circuit[tsim] (see this issue for details: https://github.com/QuEraComputing/bloqade-circuit/issues/837)

And I think we can't use _tsim_import_error directly because the variable will be cleared after the "except" block, so we have to store it to be persistent in the class if we want lazy imports. I made a change to store the _tsim_import_error as a field in the class instead of defining a global variable.

david-pl: > Yeah, I agree that it is rendundant, but the problem is that the import error isn't always from not having installed bloqade-circuit[tsim]

Yes, I understand. What I meant is that both `raise from` _and_ the error message now print the original ipywidgets error now.

> And I think we can't use _tsim_import_error directly because the variable will be cleared after the "except" block, so we have to store it to be persistent in the class if we want lazy imports.

I don't understand, everything is inside the `except` block, no? But I'm admittedly not too familiar with lazy imports.

In any case, these are just nits. Feel free to merge.

david-pl (approved):

src/bloqade/tsim/circuit.py
jasonhan3: > Yes, I understand. What I meant is that both raise from and the error message now print the original ipywidgets error now.

I think the error message is 

`ImportError: tsim is required for bloqade.tsim.Circuit. Install with: pip install "bloqade-circuit[tsim]". Tsim import error: No module named 'ipywidgets'`
So the 'ipywidgets' message will only pop up once? But admittedly, it does look kind of awkward

> I don't understand, everything is inside the except block, no? But I'm admittedly not too familiar with lazy imports.

The class is defined in the `except` block, but the class is used later in the code outside of the except block. So if the class is used outside the block, then `_tsim_import_error` won't be accessible outside the block, so we need to store it somehow I think.

PR #840, opened 2026-07-08
feat: added helper kernel for shorthand input format

jasonhan3:
Alternative solution to address https://github.com/QuEraComputing/bloqade-circuit/issues/341 .

Basically, we define a helper kernel that takes in a list of tuples of (str, float) that specify the pauli products and probabilities of applying the pauli products.

Example:

two_qubit_pauli_channel_shorthand([("XX", 0.1), ("YI", 0.2), ("IZ": 0.3)], control, target)

However, we don't do validation in the kernel, so the following edge cases will not error but will have implementation-specific behavior:

- duplicate strings in the probabilities list (seeing "XX" twice)
- length > 15
- strings that lie outside of {IX, IY, IZ, XI, XX, XY, XZ, YI, YX, YY, YZ, ZI, ZX, ZY, ZZ}

### DIFF
--- a/src/bloqade/squin/__init__.py
+++ b/src/bloqade/squin/__init__.py
@@ -1,3 +1,5 @@
+"""Various standard library functions re-exported."""
+
 from bloqade.decoders.dialects import annotate as annotate
 from bloqade.decoders.dialects.annotate import (
     set_detector as set_detector,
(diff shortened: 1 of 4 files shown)


jasonhan3:
Closed; not our desired solution

PR #841, opened 2026-07-09
feat: added draft for implementation with IR manipulation

jasonhan3:
Alternative implementation for https://github.com/QuEraComputing/bloqade-circuit/issues/331 that manipulates the IR, and also supports kernels where qubits and theta are interleaved.

For example, for the kernel

```
@squin.kernel
def another_kernel(theta: float, q0: Qubit, phi: float, q1: Qubit) -> None:
...
```

calling `wrapped_kernel = squin.wrap(another_kernel, 0.25, 0.5)` is the same as

```
@squin.kernel
def wrapped():
    qs = squin.qalloc(2)
    kernel(0.25, qs[0], 0.5, qs[1])
    return squin.qubit.broadcast.measure(qs)
```

### DIFF
--- a/src/bloqade/squin/__init__.py
+++ b/src/bloqade/squin/__init__.py
@@ -1,3 +1,5 @@
+"""Re-export imports in public API."""
+
 from bloqade.decoders.dialects import annotate as annotate
 from bloqade.decoders.dialects.annotate import (
     set_detector as set_detector,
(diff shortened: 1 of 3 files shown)


jasonhan3:
Closed; the overall scaffolding is too complex for the feature, and also, this PR has bugs to fix.

PR #843, opened 2026-07-09
Implementation with two lists for syntactic sugar for two qubit pauli channel

jasonhan3:
Define a custom lowering pass (`FromPythonCallTwoQubitPauliChannelSimple`) on a new statement (`TwoQubitPauliChannelSimple`) that lowers

```
squin.two_qubit_pauli_channel(paulis, probs, q[0], q[1])
```

to a `TwoQubitPauliChannel`.

The `FromPythonCallTwoQubitPauliChannel` custom lowering has two responsibilities (because it supports the 'broadcast'ed version of the statement AND the syntactic sugar):
1. Checks if the 'control'/'target' argument is a single qubit or a list of qubits.
2. Checks if the user passes one argument (`probabilities`) or two arguments (`paulis`, `probabilities`) and parses the latter "syntactic sugar" case accordingly.

Also attempted an implementation for the broadcasted version.

Due to the broadcasted version, for the `TwoQubitPauliChannel` statement, changed the typevar for the length for the `controls` and `targets` iList to be `Any` instead of `N`, but added check() statements to check the static lengths.

### DIFF
--- a/src/bloqade/cirq_utils/emit/noise.py
+++ b/src/bloqade/cirq_utils/emit/noise.py
@@ -44,6 +44,7 @@ class __EmitCirqNoiseMethods(MethodTable):
         p = frame.get(stmt.p)
         controls = frame.get(stmt.controls)
         targets = frame.get(stmt.targets)
+        self.check_two_qubit_lengths(controls, targets)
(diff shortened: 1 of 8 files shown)


jasonhan3:
Closed, as this solution is too complex for the use case

PR #845, opened 2026-07-15
Add Lost-aware XOR for measurement results

simsaidan:
Allow meas[0] ^ meas[1] outside and inside kernels, returning Lost if either operand is Lost.

### DIFF
--- a/src/bloqade/qubit/_typeinfer_binop.py
+++ b/src/bloqade/qubit/_typeinfer_binop.py
@@ -1,0 +1,15 @@
+"""Type inference overlay for BitXor on MeasurementResult."""
+
+from kirin import interp
+from kirin.dialects import py
+from kirin.dialects.py.binop.typeinfer import TypeInfer
+
+from bloqade.types import MeasurementResultType
+
+
+@interp.impl(py.BitXor, MeasurementResultType, MeasurementResultType)
+def bitxor_measurement(self, interp_, frame, stmt):
+    return (MeasurementResultType,)
+
+
+TypeInfer.bitxor_measurement = bitxor_measurement
--- a/src/bloqade/types.py
+++ b/src/bloqade/types.py
@@ -12,7 +12,64 @@ from kirin.dialects.ilist import IList
 from bloqade.decoders.dialects.annotate.types import (
     MeasurementResult as MeasurementResult,
     MeasurementResultType as MeasurementResultType,
+    MeasurementResultValue as MeasurementResultValue,
 )
+
+
+def _as_measurement_value(
+    other: Any,
+) -> MeasurementResultValue | None:
+    if isinstance(other, MeasurementResult):
+        return other.value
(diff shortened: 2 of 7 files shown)


src/bloqade/types.py
```diff
@@ -12,9 +12,66 @@
...
+    lhs = _as_measurement_value(other)
+    if lhs is None:
+        return NotImplemented
+    return MeasurementResult(_xor_measurement_values(lhs, self.value))
+
+
+MeasurementResultValue.__xor__ = _measurement_value_xor  # type: ignore[method-assign]
+MeasurementResultValue.__rxor__ = _measurement_value_rxor  # type: ignore[method-assign]
+MeasurementResult.__xor__ = _measurement_result_xor  # type: ignore[method-assign]
+MeasurementResult.__rxor__ = _measurement_result_rxor  # type: ignore[method-assign]
```
david-pl, line 72: Right, I just realized the types moved to bloqade-decoders. We should move this PR instead of assigning additional methods here.

src/bloqade/qubit/_typeinfer_binop.py
```diff
@@ -0,0 +1,15 @@
...
+
+from bloqade.types import MeasurementResultType
+
+
+@interp.impl(py.BitXor, MeasurementResultType, MeasurementResultType)
+def bitxor_measurement(self, interp_, frame, stmt):
+    return (MeasurementResultType,)
+
+
+TypeInfer.bitxor_measurement = bitxor_measurement
```
david-pl, line 15: Please don't add methods like this. It should be possible to register a new method with a method table, though I admit I'm not 100% sure since there already is an impl for `py.BitXor` somewhere. Worst-case, we can inherit from the existing method table and override the method.

src/bloqade/types.py
```diff
@@ -12,9 +12,66 @@
...
+    if isinstance(other, MeasurementResultValue):
+        return other
+    return None
+
+
+def _xor_measurement_values(
+    lhs: MeasurementResultValue, rhs: MeasurementResultValue
+) -> MeasurementResultValue:
+    if lhs is MeasurementResultValue.Lost or rhs is MeasurementResultValue.Lost:
+        return MeasurementResultValue.Lost
```
david-pl, line 33: This is an important semantic decision: should any be `Lost`, the XOR will always return `Lost`. This needs to be documented somewhere.

david-pl (changes requested):
@simsaidan  Thanks for the work! Overall, I like the shape, but realized this should move into bloqade-decoders.

Also, what happens if you do e.g. `MeasurementResult.One ^ 1`? I think the `IntEnum` falls back to standard XOR semantics then. Could we prevent that?

simsaidan:
@david-pl thanks for the feedback! I moved some of this pr over to https://github.com/QuEraComputing/bloqade-decoders/pull/50. Let me know if that is what you had in mind.

PR #849, opened 2026-07-24
Add leakage to squin and stim

david-pl:


pyproject.toml
```diff
@@ -14,7 +14,7 @@ dependencies = [
     "bloqade-decoders>=0.4.0,<0.7.0",
     "numpy>=1.22.0",
     "scipy>=1.13.1",
-    "kirin-toolchain~=0.22.2",
```
kaihsin, line 17: I would suggest soft pin and set lower bound to 0.22.11 instead of hard pin

kaihsin (changes requested):

johnzl-777 (approved):

johnzl-777:
Not the happiest with what I had to do to get the CI to be happy here, added some rule exclusions because the CI always pulls in the latest version of ruff and I'd have to add a bunch of new docstrings (can be done in a separate PR, I've avoided that for now due to the number needed)

PR #857, opened 2026-08-31
feat: accumulated detector/observable ID's during MeasurementIDAnalysis

### LINKED ISSUE #856: MeasurementIDAnalysis: accumulate detectors and observables, not just return them
## Summary

`MeasurementIDAnalysis` mints a `DetectorId` / `ObservableId` for every `annotate.SetDetector` / `SetObservable` it visits, but only returns it into the frame. Nothing accumulates them, so a consumer can only find a detector that happens to appear in the kernel's **return value**.

Please also collect them on the analysis, so consumers can enumerate every detector and observable in a program regardless of what the kernel returns.

## Where it stands today

`bloqade/analysis/measure_id/impls.py` already does the hard part — it handles both statements and assigns ordinals:

```python
@interp.impl(annotate.stmts.SetDetector)
def set_detector(self, interp_, frame, stmt):
    frame.num_measures_at_stmt[stmt] = interp_.measure_count
    detector_value = DetectorId(idx=interp_.detector_count, data=frame.get(stmt.measurements))
    interp_.detector_count += 1
    return (detector_value,)          # <- returned, never collected
```

`MeasurementIDAnalysis` already carries `detector_count` / `observable_count` and resets them in `initialize`, and `lattice.DetectorId` / `ObservableId` already carry an `idx`. The only missing piece is the list.

## Proposed change

Add two lists alongside the existing counters and append in the two impls:

```python
class MeasurementIDAnalysis(ForwardExtra[MeasureIDFrame, MeasureId]):
    def initialize(self) -> Self:
        self.measure_count = 0
        self.detector_count = 0
        self.observable_count = 0
        self.detectors: list[DetectorId] = []       # new
        self.observables: list[ObservableId] = []   # new
        return super().initialize()
```

and in each impl, `interp_.detectors.append(detector_value)` before returning. `idx` already gives a stable order.

(Assigned in `initialize` rather than as class attributes, since the existing counters are class attributes and mutable defaults would be shared across instances.)

Purely additive — no existing behaviour changes.

## Why we want it

In bloqade-lanes we have two post-processing implementations that must agree, and this is the last thing keeping the older one alive:

- `bloqade.gemini.post_processing.generate_post_processing` builds on `MeasurementIDAnalysis` and reconstructs the kernel's return value. It has no detector/observable emitters, because it can't enumerate them.
- `bloqade.lanes.analysis.atom.AtomInterpreter.get_post_processing` walks the *lowered* kernel and does collect them, in its own side channel:

  ```python
  @interp.impl(annotate.stmts.SetDetector)
  def set_detector(self, interp_, frame, stmt):
      result = DetectorResult(frame.get(stmt.measurements))
      interp_._detectors.append(result)     # the side channel this issue is about
      return (result,)
  ```

The two already produce identical user values, and the older one is deprecated for that purpose. But it can't be removed while it's the only source of detectors and observables, so downstream currently assembles a hybrid: user values from `generate_post_processing`, detectors and observables from the atom analysis.

That matters beyond tidiness. The atom analysis indexes measurements by a record id assigned over the *lowered* IR, which only lines up with `MeasurementIDAnalysis`'s numbering while lowering preserves the ordering — an invariant we maintain by hand and pin with a test. Sourcing everything from `MeasurementIDAnalysis` removes that coupling from the pipeline.

Downstream tracking issue: QuEraComputing/bloqade-lanes#981 (which currently proposes a subclass on our side; doing it here instead avoids duplicating the impl table and having it drift).

## Note for whoever picks this up

Downstream reduces a detector as `bool(np.logical_xor.reduce(values, axis=0))` rather than `functools.reduce(operator.xor, ...)`. The two differ on an annotation covering no measurements: the former is `False`, the latter raises `TypeError`. If emitters land here too, worth matching deliberately.

jasonhan3:
closes https://github.com/QuEraComputing/bloqade-circuit/issues/856

- added detectors and observables tracking to MeasurementIDAnalysis
- helps with https://github.com/QuEraComputing/bloqade-lanes/issues/981

## Summary

Updates `MeasurementIDAnalysis` to retain detector and observable IDs encountered during analysis.

- Adds `detectors` and `observables` collections to the analysis result.
- Records each `DetectorId` and `ObservableId` when processing `set_detector` and `set_observable` statements.
- Resets these collections in `initialize()`, allowing an analysis instance to be reused without retaining results from a previous run.
- Adds Ruff-required public API docstrings for the affected measure-ID analysis interfaces.

## Testing

```console
pytest test/analysis/measure_id/test_measure_id.py
# 15 passed

pre-commit run --files .pre-commit-config.yaml \
  src/bloqade/analysis/measure_id/analysis.py \
  src/bloqade/analysis/measure_id/impls.py
# all hooks passed
```

### DIFF
--- a/src/bloqade/analysis/measure_id/analysis.py
+++ b/src/bloqade/analysis/measure_id/analysis.py
@@ -6,15 +6,18 @@ from kirin.analysis import ForwardExtra
 from typing_extensions import Self
 from kirin.analysis.forward import ForwardFrame
 
-from .lattice import MeasureId, NotMeasureId
+from .lattice import MeasureId, DetectorId, NotMeasureId, ObservableId
 
 
 @dataclass
 class MeasureIDFrame(ForwardFrame[MeasureId]):
+    """Analysis frame that tracks measurement counts for each statement."""
+
     num_measures_at_stmt: dict[ir.Statement, int] = field(default_factory=dict)
 
 
 class MeasurementIDAnalysis(ForwardExtra[MeasureIDFrame, MeasureId]):
+    """Forward analysis that assigns and propagates measurement identifiers."""
 
     keys = ["measure_id"]
     lattice = MeasureId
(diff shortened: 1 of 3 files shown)


david-pl (approved):
LGTM

PR #859, opened 2026-09-04
WIP: circuit depth validation

david-pl:


src/bloqade/analysis/count_statements/impls.py
```diff
@@ -0,0 +1,140 @@
...
+def _resolve_callable(fn: ir.SSAValue):
+    mt = CountStatementAnalysis.maybe_const(fn, ir.Method)
+    if mt is not None:
+        return mt.code
+
+    owner = fn.owner
+    if isinstance(owner, (func.Lambda, func.Function)):
+        return owner
+
+    return None
```
weinbe58: I believe you can use a trait to get the callable region instead of dispatching on a particular statement. 

https://github.com/QuEraComputing/kirin/blob/395271c3ca7edced4eb505cad6d5be14130152a6/src/kirin/ir/traits/callable.py#L17

I would do this so that if someone extends the function dialect this analysis will still work

src/bloqade/analysis/count_statements/analysis.py
```diff
@@ -0,0 +1,82 @@
...
+    ```
+    """
+
+    keys = ("count.statements",)
+    lattice = EmptyLattice
+
+    predicate: Callable[[Statement], tuple[bool, int, int]]
+    """``(matched, idx, increment)`` for each visited statement."""
+    N: int = 1
+    """Number of counter buckets in ``counts``."""
```
weinbe58: why not
```suggestion
    keys = ("count.statements",)
    lattice = EmptyLattice

    predicate: Callable[[Statement], tuple[bool, int]]
    """``(matched, idx, increment)`` for each visited statement."""
    counter: Counter[Statement]
```

Why not use a Counter for this? The predicate returns just returns `bool` to and the increment.

weinbe58 (commented):
Looks OK, it seems a little odd to have this counter set up ahead of time. it seems error prone compared to just using a Counter object.

src/bloqade/analysis/count_statements/analysis.py
david-pl: I like this idea in principle, but it's not that easy. The problem is that `Counter[Statement]` depends on the hashes, rather than just the type.

So, counting to stdlib calls, e.g.

```python
@squin.kernel
def main():
    q = squin.qalloc(2)
    squin.x(q[0])
    squin.x(q[1])
```

would report a counter that is `{X(): 2}` since it's the same statement. However, inlining the above kernel you'd get `{X(): 1, X(): 1}` since now there are two `X` statements with different IDs.

david-pl: To solve this, I first went for an approach to use `counter: Counter[type]` and then increment as `counter[type(node)] += increment`.

However, this is also not good since the `CircuitDepthValidation` needs to aggregate the total number of single-qubit and two-qubit gates whenever checking the threshold. Finally, I ended up with a design where you now also return a configurable key for the counter

```python
CounterKey = TypeVar("CounterKey", bound=Hashable)

class CountStatementsAnalsyis(Generic[CounterKey])
    predicate: Callable[[Statement], tuple[CounterKey, int] | None]
    counter: Counter[CounterKey]
```

And for the `CircuitDepthValidation`, I used

```python
class _GateKind(enum.Enum):
    SINGLE_QUBIT_GATE = enum.auto()
    TWO_QUBIT_GATE = enum.auto()


def _count_gates(node: ir.Statement):
    if isinstance(node, SingleQubitGate):
        return _GateKind.SINGLE_QUBIT_GATE, 1
    elif isinstance(node, TwoQubitGate):
        return _GateKind.TWO_QUBIT_GATE, 1
```

weinbe58 (approved):
LGTM, thanks! don't forget to backport this!

PR #861, opened 2026-09-15
jeff: dialect

nathanieltornow:
The jeff exchange format (https://github.com/unitaryfoundation/jeff) as a kirin dialect, ~~including verification~~.
First of a series that adds conversion from squin to jeff and back, so squin kernels can be exchanged with other compilers.


src/bloqade/jeff/dialects/stmts/_analysis.py
```diff
@@ -0,0 +1,75 @@
...
+from bloqade.jeff.types import WireType
+
+from .call import Return
+from .quantum import Ppr, Gate
+from ._dialect import dialect
+from .classical import ConstInt, ConstFloat
+
+
+@dialect.register(key="typeinfer")
+class TypeInfer(interp.MethodTable):
```
david-pl: Why do we need custom type inference methods here? I think kirin should be able to handle this given the typed fields.

src/bloqade/jeff/dialects/stmts/_analysis.py
```diff
@@ -0,0 +1,75 @@
...
+        else:
+            output = types.Generic(tuple, *values)
+        return interp.ReturnValue(output)
+
+
+@dialect.register(key="constprop")
+class ConstProp(interp.MethodTable):
+    @interp.impl(ConstInt)
+    @interp.impl(ConstFloat)
+    def constant(
```
david-pl: Similarly here: I'm surprised we need this. Shouldn't setting the `ConstantLike` trait take care of this? Might be caused by lack of a concrete interpreter for Jeff though.

src/bloqade/jeff/dialects/group.py
```diff
@@ -0,0 +1,19 @@
...
+
+from collections.abc import Callable
+
+from kirin import ir
+from kirin.dialects import func, ssacfg
+
+from .stmts import dialect
+
+
+@ir.dialect_group([dialect, func, ssacfg])
```
david-pl: I think it would make sense to break this jeff dialect down into multiple smaller dialects. You're essentially defining a single dialect for the entire jeff dialect group. For example, you could split this into:

* classical
* wire (allocate, free, reset, measure, etc.)
* quantum (basically just the gate statement, I think)
* scf

Not sure about `call` here since, if I understand correctly, it needs to be composed with kirin's `func`. Still, it might make sense to define a jeff `func` dialect and then compose this in a smaller group with kirin's `func`.

The overall `jeff` dialect group would compose all of these together.

src/bloqade/jeff/dialects/group.py
```diff
@@ -0,0 +1,19 @@
...
+def kernel(self) -> Callable[[ir.Method], None]:
+    """The group of pure jeff programs: the jeff dialect, `func`, and `ssacfg`."""
+
+    def run_pass(method: ir.Method) -> None:
+        """Run nothing; the group defines no passes of its own."""
```
david-pl: Should we run some validation here?

src/bloqade/jeff/dialects/region_utils.py
```diff
@@ -0,0 +1,41 @@
+"""Containment questions about kirin IR."""
+
+from collections.abc import Iterator
+
+from kirin import ir
+
+
+def ancestors(node: ir.Statement) -> Iterator[tuple[ir.Block, ir.Region]]:
```
david-pl: I think you re-invented some of kirin's IR walking utilities here. This can probably be removed and calls replaced by `IRNode.is_ancestor`.

I think you can also use it for the two methods below so you could remove this whole file.

src/bloqade/jeff/py.typed
david-pl, line 1: What's this file?

src/bloqade/jeff/verify/__init__.py
david-pl, line 1: Verification is extensive, which is good. But without going into detail now, I think these checks should be refactored. Validation should be done using kirin's `ValidationPass` and `ValidationSuite` together with appropriate analyses runs.

`Report` kind of re-invents that and the `JeffVerificationError` should essentially just be a `ir.ValidationError`. Using analysis runs also saves you the trouble of going through extensive `if isinstance` checks.

david-pl (changes requested):
Overall, the changes here are solid. I just have some small comments and two larger structural ones:

* The jeff dialect could use splitting up into smaller dialects, which you then compose into a single jeff dialect group (similar to how e.g. stim has multiple dialects even though it's a single target).
* The verification should be done using kirin's validation passes.

src/bloqade/jeff/dialects/stmts/_analysis.py
nathanieltornow: I added them since kirin was not doing correct type inference on variadic types. But removed them now, and will bring them back potentially when we need to do passes on jeff dialect itself

src/bloqade/jeff/dialects/stmts/_analysis.py
nathanieltornow: Also removed for now

src/bloqade/jeff/dialects/group.py
nathanieltornow: done 👍

src/bloqade/jeff/dialects/group.py
nathanieltornow: done 👍

src/bloqade/jeff/dialects/region_utils.py
nathanieltornow: thanks, I use the kirin internals now, removed this file

src/bloqade/jeff/py.typed
nathanieltornow: removed

src/bloqade/jeff/verify/__init__.py
nathanieltornow: did split this PR now and will do verification using the ValidationPass and ValidationSuite in the next PR to keep this PR smaller

src/bloqade/jeff/dialects/stmts/__init__.py
```diff
@@ -0,0 +1,107 @@
...
+    FloatArraySet as FloatArraySet,
+    IntArrayConst as IntArrayConst,
+    FloatArrayZero as FloatArrayZero,
+    FloatPredicate as FloatPredicate,
+    IntArrayCreate as IntArrayCreate,
+    FloatArrayConst as FloatArrayConst,
+    FloatArrayCreate as FloatArrayCreate,
+)
+
+DIALECTS = (classical.dialect, wire.dialect, gate.dialect, scf.dialect, call.dialect)
```
david-pl: Unused. Remove?

src/bloqade/jeff/dialects/group.py
```diff
@@ -0,0 +1,49 @@
...
+    """Define the dialect group for pure jeff programs.
+
+    The group holds every jeff dialect, kirin's `func` dialect and kirin's `ssacfg`
+    dialect.
+    """
+
+    def run_pass(method: ir.Method) -> None:
+        """Run kirin's structure checks and type checks on the method."""
+        method.verify()
+        method.verify_type()
```
david-pl: Is typeinfer run anywhere? Otherwise, this may lead to issues.

src/bloqade/jeff/signature.py
```diff
@@ -0,0 +1,31 @@
...
+
+def function_of(method: ir.Method[..., Any]) -> func.Function:
+    """Return the function that a kernel method wraps.
+
+    Raises:
+        TypeError: If the method does not wrap a function.
+    """
+    code = method.code
+    if not isinstance(code, func.Function):
+        raise TypeError(f"'{method.sym_name}' is not a function: {code.name}")
```
david-pl: Do we really need this helper? It's essentially just accessing `method.code`. Not sure if we need the error at this point either.

src/bloqade/jeff/types.py
```diff
@@ -0,0 +1,118 @@
...
+    """
+    return kind.is_subseteq(WireType) or kind.is_subseteq(QuregType)
+
+
+def is_bit(value: ir.SSAValue) -> bool:
+    """Check whether a value is a jeff 1-bit integer.
+
+    Kirin types a jeff bit as `Bool`.
+    """
+    return value.type.is_subseteq(types.Bool)
```
david-pl: FYI, `is_subseteq` can be a bit of footgun: this will return `True` for values that are `BottomType`.  You may want to explicitly check for that to avoid issues caused by type inference failure, which will be difficult to debug because of this.

david-pl (commented):
Looks good! Just some small comments.

src/bloqade/jeff/signature.py
nathanieltornow: removed

src/bloqade/jeff/dialects/group.py
nathanieltornow: removed the method now fully, since we don't need to have a frontend for jeff anyways

src/bloqade/jeff/dialects/stmts/__init__.py
nathanieltornow: removed

src/bloqade/jeff/types.py
nathanieltornow: added explicit helper

david-pl (approved):

PR #862, opened 2026-09-17
jeff: validation

nathanieltornow:
Adds kirin ValidationPasses that check jeff IR.

**Structure**:
  - Runs verify_type.
  - Checks that each function body is one block that ends in a jeff return.
  - Rejects non-jeff statements and statements after a terminator.
  - Checks operand scope, and checks that functions and For, Switch and While regions are isolated. Those three statements now declare IsolatedFromAbove.

**LinearityValidation**: checks that each function uses each wire and register exactly once.


src/bloqade/jeff/analysis/validation/__init__.py
```diff
@@ -0,0 +1,50 @@
...
+from kirin.validation import ValidationPass
+
+from bloqade.jeff.dialects import kernel
+
+from .base import CheckFrame
+from .linearity import LinearityAnalysis
+from .structure import StructureAnalysis
+
+
+class StructureValidation(ValidationPass[CheckFrame[EmptyLattice]]):
```
david-pl: Just a style preference on my end: I think code should live in dedicated files. `__init__.py` should be reserved for exports only.

src/bloqade/jeff/analysis/validation/__init__.py
```diff
@@ -0,0 +1,50 @@
...
+    def run(
+        self, method: ir.Method
+    ) -> tuple[CheckFrame[EmptyLattice], list[ir.ValidationError]]:
+        """Run the linearity analysis and return its frame and validation errors."""
+        analysis = LinearityAnalysis(kernel)
+        frame, _ = analysis.run(method)
+        return frame, analysis.get_validation_errors()
+
+
+PASSES: list[type[ValidationPass[CheckFrame[EmptyLattice]]]] = [
```
david-pl: One-time use in a test. Can we remove that?

src/bloqade/jeff/analysis/validation/base.py
```diff
@@ -0,0 +1,173 @@
...
+from bloqade.jeff.dialects import stmts
+
+SHARED_KEY = "validate.jeff"
+"""The registry key of the rules that all analyses share."""
+
+L = TypeVar("L", bound=lattice.BoundedLattice)
+
+
+@dataclass
+class CheckFrame(ForwardFrame[L]):
```
david-pl: Can you explain a bit more why we need this custom frame class?

src/bloqade/jeff/analysis/validation/base.py
```diff
@@ -0,0 +1,173 @@
...
+    """A dataflow frame that reports each missing operand to its analysis."""
+
+    check: "Check[L]" = field(kw_only=True)
```
david-pl: I will say that a frame that references the analysis is a bit odd. I feel like this might lead to cyclic issues, since interpreters also own the frames.

src/bloqade/jeff/analysis/validation/base.py
```diff
@@ -0,0 +1,173 @@
...
+
+    def method_self(self, method: ir.Method) -> L:
+        """Return lattice top for the method object."""
+        return self.lattice.top()
+
+    def error(self, node: ir.IRNode, message: str) -> None:
+        """Record a validation error at an IR node."""
+        self.add_validation_error(node, ir.ValidationError(node, message))
+
+    def enter_function(self, method: ir.Method) -> bool:
```
david-pl: Is it possible to instead override `interp.frame_call` or similar?

src/bloqade/jeff/analysis/validation/base.py
```diff
@@ -0,0 +1,173 @@
...
+        self, frame: CheckFrame[L], node: ir.Statement, value: ir.SSAValue
+    ) -> None:
+        """Handle an operand that has no entry in the current frame.
+
+        The default does nothing.
+        """
+
+    def frame_eval(
+        self, frame: CheckFrame[L], node: ir.Statement
+    ) -> interp.StatementResult[L]:
```
david-pl, line 116: Isn't this pretty much the same as the abstract method? Why do we need this override?

src/bloqade/jeff/analysis/validation/base.py
```diff
@@ -0,0 +1,173 @@
...
+
+
+@dataclass
+class Check(ForwardExtra[CheckFrame[L], L], ABC):
+    """A forward analysis that records validation errors at IR nodes."""
+
+    isolated: ClassVar[bool] = False
+    """If true, each region of an isolated statement runs in a fresh frame."""
+
+    visited: set[ir.Statement] = field(default_factory=set, init=False)
```
david-pl: Shouldn't forward interpretation guarantee we only hit each statement once? Also, could it be that this skips over functions even when called with different arguments? If so, then you might validate a function call with invalid arguments because it was validated before.

src/bloqade/jeff/analysis/validation/linearity.py
```diff
@@ -0,0 +1,40 @@
...
+    def leave_function(self, frame: CheckFrame[EmptyLattice]) -> None:
+        """Check the use count of every wire and register that the function defines.
+
+        The count includes only uses by statements inside the function.
+        """
+        code = frame.code
+        for value in frame.entries:
+            if not is_linear(value.type):
+                continue
+            uses = sum(1 for use in value.uses if code.is_ancestor(use.stmt))
```
david-pl:
```suggestion
            uses = sum(code.is_ancestor(use.stmt) for use in value.uses)
```

src/bloqade/jeff/analysis/validation/structure.py
```diff
@@ -0,0 +1,111 @@
...
+    """Report a terminator that another statement follows in its block."""
+    block = terminator.parent_block
+    if block is not None and terminator is not block.last_stmt:
+        check.error(terminator, "a block has statements after its terminator")
+
+
+def _owner_name(block: ir.Block) -> str:
+    """Return the name of the statement that owns a block."""
+    owner = block.parent_stmt
+    return owner.name if owner is not None else "?"
```
david-pl: This doesn't deserve a dedicated helper IMO.

src/bloqade/jeff/dialects/group.py
```diff
@@ -5,23 +5,22 @@
 
 from .stmts import scf, call, gate, wire, classical
 
+DIALECTS: tuple[ir.Dialect, ...] = (
```
david-pl: That came back. Why? 😅

david-pl (commented):
Looks good overall, just some minor comments. One thing I'm not sure about is whether we really need that custom frame type.

src/bloqade/jeff/analysis/validation/__init__.py
nathanieltornow: Moved into "passes.py"

src/bloqade/jeff/analysis/validation/__init__.py
nathanieltornow: removed

src/bloqade/jeff/analysis/validation/base.py
nathanieltornow: I had the custom frame because `Frame.get` raises when a statement reads a value not in the frame, so we can't continue to collect more errors. 

Changed to plain `Forward` now, with checking `frame.entries` before reading the values

src/bloqade/jeff/analysis/validation/base.py
nathanieltornow: true, removed now

src/bloqade/jeff/analysis/validation/base.py
nathanieltornow: done

src/bloqade/jeff/analysis/validation/base.py
nathanieltornow: the `frame_eval` from kirin raises when a statement is in a dialect not in the group. We pass it into a fallback to record this error.

src/bloqade/jeff/analysis/validation/structure.py
nathanieltornow: inlined

src/bloqade/jeff/dialects/group.py
nathanieltornow: i added it to now do the check "if all statements are in the dialect group", but removed it now again 👍

src/bloqade/jeff/analysis/validation/base.py
nathanieltornow: We need it to prevent crash on recursion. However, I missed the fact that one could validated a function that has invalid arguments. So, now we key them by function and arguments, such that only a function that has been called with the exact arguments is skipped.

nathanieltornow:
Thanks @david-pl, all issues should be resolved

src/bloqade/jeff/analysis/validation/base.py
david-pl: I see. I'm not sure this is ideal. If kirin treats this as a hard failure, why should we fall back to a ValidationError here? Also, `mt.verify` will error on foreign dialect statements, I think, so usually you shouldn't see this during validation. Not sure though. Definitely not ideal for validation, however. Maybe something to bring up on the kirin repo itself.

src/bloqade/jeff/analysis/validation/base.py
```diff
@@ -0,0 +1,188 @@
...
+        self, frame: ForwardFrame[L], node: ir.Statement, values: Iterable[ir.SSAValue]
+    ) -> tuple[L, ...]:
+        """Return the lattice values of the operands `values` of `node`.
+
+        For an operand that the frame does not hold, the method calls
+        `operand_missing` and stores lattice top.
+        """
+        values = tuple(values)
+        for value in values:
+            if value not in frame.entries:
```
david-pl: This, combined with `frame.get_values` iterates over `entries` twice. You can also just do `frame.entries.get(value, self.lattice.top())` and accumulate the result in the loop.

david-pl (approved):
Just one comment, and that's not a blocker. Solid PR!

Issue #529, opened 2025-10-07
`QuantumState use case`: variation distance


jon-wurtz:
Given a quantum state or probability distribution, and another quantum state, probability distribution, or set of frequentist samples, compute the variation distance (cross entropy, JS, KL divergence, Bhattacharyya distance) between the distributions.

Use cases: Actively used to diagnose problems and noise on Aquila, including quantum reservoir computing. Further, used in ML inference of noise, benchmarking tasks.

yzcj105:
Adding one more to the list: Total Variation Distance

jon-wurtz:
As an explicit code snippet,

```python
data = get_bitstrings_from_gemini()
state1 = get_first_guess_circuit_state()
state2 = get_other_guess_circuit_state()
distance1 = state1.js_divergence(data)
distance2 = state2.js_divergence(data)
if distance1<distance2:
    print("First guess is more likely")
else:
    print("Second guess is more likely")

x_ = np.linspace(0,1,100)
distance = [get_state(param = x).kl_divergence(data) for x in x_]
plt.plot(x_, distance)
# Look at where the distance is minimized to identify most likely models
```

PR #863, opened 2026-09-17
feat: added tomography result class

jasonhan3:
- Added a class for getting Tomography from shot counts.

Wanted to create a PR to start the discussion of adding a `TomographyResult` class, in order to reconstruct a quantum state from basis measurements

I think it's related to this issue: https://github.com/QuEraComputing/bloqade-circuit/issues/529


src/bloqade/analysis/tomography/tomography.py
```diff
@@ -0,0 +1,122 @@
...
+
+from __future__ import annotations
+
+import math
+from dataclasses import dataclass
+from collections.abc import Mapping, Sequence
+
+import numpy as np
+
+BASES = ("X", "Y", "Z")
```
david-pl: You convert this to a `set` a few times and all other usage, as far as I can tell, would also work if you just made this a `set`.

src/bloqade/analysis/tomography/tomography.py
```diff
@@ -0,0 +1,122 @@
...
+
+
+def _density_matrix_from_bloch(bloch: Mapping[str, float]) -> np.ndarray:
```
david-pl, line 14: Hm, this is tricky: mixed states are not normalized, but if you leave things unnormalized then you can get a bloch vector with norm > 1 leading to invalid density matrices:

```python
import numpy as np
from bloqade.analysis.tomography import TomographyResult
shots = {basis: np.array([0]) for basis in ("X", "Y", "Z")}
result = TomographyResult(shots)
print(np.linalg.eigvalsh(result.density_matrix))  # negative eigenvalue is not valid
```

A solution to fix the above would be to normalize if `norm > 1`, but that doesn't guarantee that the norm of a Bloch vector for a mixed state is too large so long as it's < 1.

src/bloqade/analysis/tomography/tomography.py
```diff
@@ -0,0 +1,122 @@
...
+) -> float:
+    overlap = float(np.real(np.trace(density_matrix @ target_density_matrix)))
+    det_product = float(
+        np.real(np.linalg.det(density_matrix))
+        * np.real(np.linalg.det(target_density_matrix))
+    )
+    return overlap + 2.0 * math.sqrt(max(det_product, 0.0))
+
+
+@dataclass(frozen=True, init=False)
```
david-pl: Having a frozen dataclass with a mutable field (you can mutate the density matrix, _and_ you set it in the `__init__`) and a custom `__init__` is odd, I think. I'd suggest to at least remove the `frozen=True`.  Might also make sense to make the `__init__` a `__post_init__`. Or, maybe even don't make it a dataclass at all if you need the `__init__`?

david-pl (changes requested):
There's a tricky issue with normalization, and the frozen dataclass feels off.

src/bloqade/analysis/tomography/tomography.py
jasonhan3: I see, yeah this is tricky. I was looking into other quantum computing libraries, and a common pattern is to have some kind of "fitter" that finds the most likely physically valid density matrix given the measurement outcomes (examples: [Forest-Benchmarking](https://forest-benchmarking.readthedocs.io/en/latest/examples/tomography_state.html), [Qiskit](https://qiskit-community.github.io/qiskit-experiments/manuals/verification/state_tomography)). We could do something similar, and have a `DensityMatrix` object instead? And add methods for computing fidelity to other states to that class (such as `def fidelity_bloch(target_bloch: np.ndarray)`).

To implement this `DensityMatrix` object, we could probably reuse the existing `QuantumState` class (https://github.com/QuEraComputing/bloqade-circuit/blob/7b56a8f734897febd11c23397065894713278f69/src/bloqade/pyqrack/device.py#L28), as it seems to serve a similar purpose (to represent density matrices).

david-pl: This sounds like a good idea. I'd be happy to refactor the `QuantumState` class, but it might lead to breaking changes, which I'm not sure we'll want. Also, I'm not sure how much work this will be.

An alternative would be to document the behavior here saying that things aren't normalized for now and then leave the rest for a follow-up ticket. I'll leave that up to you.

jasonhan3: I think we can keep the `TomographyResult` class as is and define the "fitter" methods on the class as alternative constructors. And regarding interoperability with the `QuantumState` class, in a future PR we can inherit from it (won't do so now as a lot of the methods in the [`QuantumState` class](https://github.com/QuEraComputing/bloqade-circuit/blob/7b56a8f734897febd11c23397065894713278f69/src/bloqade/pyqrack/device.py#L28) are not implemented)

PR #864, opened 2026-09-18
jeff: file emit and import

nathanieltornow:
## jeff: file emit and import

Reads and writes `.jeff` files for the jeff dialects.

- `emit_jeff(method)` emits a jeff dialect kernel as a `jeff.JeffModule`. `EmitJeff` is a kirin emitter with one method table per dialect under `emit/`.
- `load_jeff(source)` reads a path, bytes or a `JeffModule` into jeff dialect IR. It runs `StructureValidation` and `LinearityValidation` and raises `JeffImportError` on a bad file.
- The new `jeff` extra installs `jeff-format`. Without the extra, the dialects and the validation still work, and `emit_jeff` and `load_jeff` raise `ImportError` with the install hint.

### Tests
- `test/jeff/programs/` holds `.jeff` files from the jeff repository, and every file must load and round-trip.
- `test/jeff/programs/rejected/` holds invalid files, and each one must fail with its expected error.
- The tests cover emit for each operation, import errors and deterministic output.


src/bloqade/jeff/emit/func.py
```diff
@@ -0,0 +1,80 @@
...
+    IntArrayType,
+    FloatArrayType,
+    is_subtype,
+    qureg_length,
+)
+
+from .base import EmitJeff, JeffFrame
+
+
+def jeff_type(kirin_type: types.TypeAttribute) -> JeffType:
```
david-pl, line 43: This looks like it should be a hard-coded `dict` and then you just get the matching element.

src/bloqade/jeff/emit/target.py
```diff
@@ -0,0 +1,38 @@
...
+from bloqade.jeff.dialects import kernel as jeff_kernel
+from bloqade.jeff.emit.base import EmitJeff
+
+
+class _Module(JeffModule):
+    """Hold a jeff module that writes its string table in sorted order."""
+
+    def _compute_strings(self) -> list[str]:
+        """Return the strings of the module in sorted order."""
+        return sorted(super()._compute_strings())
```
david-pl: Do we need this class? I wouldn't override a private method. Is this used to hook into internals? Can we make the sorting explicit inside the `emit_jeff` function below?

src/bloqade/jeff/parse/load.py
```diff
@@ -0,0 +1,55 @@
...
+    elif isinstance(source, jf.JeffModule):
+        module = source
+    else:
+        raise TypeError(
+            "load_jeff takes a path or a jeff module, and it got "
+            f"{type(source).__name__}"
+        )
+    version = module.version
+    major, minor = jf.schema.schemaVersionMajor, jf.schema.schemaVersionMinor
+    if version is None or (version.major, version.minor) != (major, minor):
```
david-pl: Should we really error on `minor` mismatch? This should probably be a semantic version check here. But I'm not sure in how far we need to make sure the versions are exactly the same.

src/bloqade/jeff/parse/lowering.py
```diff
@@ -0,0 +1,402 @@
...
+def _id(value: jf.JeffValue) -> int:
+    """Return the id of a value.
+
+    Raise `JeffImportError` if the value has no id.
+    """
+    if value.id is None:
+        raise JeffImportError(
+            "a value has no id. Refresh the module before you load it."
+        )
+    return value.id
```
david-pl: Do we need all those helpers? Effectively, you're just wrapping a one-liner with an error message. Shouldn't these errors be caught by the higher-level jeff errors anyway?

src/bloqade/jeff/parse/lowering.py
```diff
@@ -0,0 +1,402 @@
...
+    """Return the signed integer that a stored bit pattern of `bitwidth` bits encodes.
+
+    A 1-bit pattern keeps its value 0 or 1.
+    """
+    if bitwidth > 1 and pattern >= 1 << (bitwidth - 1):
+        return pattern - (1 << bitwidth)
+    return pattern
+
+
+def _kirin_type(jeff_type: jf.JeffType) -> types.TypeAttribute:
```
david-pl, line 53: Yet another case of a function that wants to be a `dict` 😄

src/bloqade/jeff/parse/lowering.py
```diff
@@ -0,0 +1,402 @@
...
+        function: jf.FunctionDef,
+        index: int | None = None,
+    ) -> ir.Method:
+        """Lower `function` to a method, and record it in `methods` under `index`.
+
+        The lowering records the method before it lowers the body, so that a
+        recursive call finds the method.
+        """
+        block = ir.Block()
+        block.args.append_from(types.Any, "self")
```
david-pl: Doesn't this make every function type `Any`?

src/bloqade/jeff/parse/lowering.py
```diff
@@ -0,0 +1,402 @@
...
+                return stmts.IntArrayCreate(tuple(inputs), bitwidth=bitwidth)
+            return stmts.FloatArrayCreate(tuple(inputs))
+        raise JeffImportError(f"unsupported {kind} op: {subkind}")
+
+    def control_flow(
+        self, state: lowering.State[Node], op: jf.JeffOp, inputs: list[ir.SSAValue]
+    ) -> ir.Statement:
+        """Return the loop or switch statement for a jeff `scf` operation."""
+        data = op.instruction_data
+        if isinstance(data, jf.ForSCF):
```
david-pl: Could use a `match` here.

david-pl (changes requested):
Looks good! Some minor comments and one thing I'd like to change: don't override a private method in the `_Module` class.

src/bloqade/jeff/emit/func.py
nathanieltornow: Done

src/bloqade/jeff/emit/target.py
nathanieltornow: It was there to enable stable hashes to test roundtrips, but I'm now comparing strings which is stable without changing the logic of jeff

src/bloqade/jeff/parse/lowering.py
nathanieltornow: yes, removed

src/bloqade/jeff/parse/lowering.py
nathanieltornow: fixed

src/bloqade/jeff/parse/lowering.py
nathanieltornow: yes, rewrote as a match, including some more locations that are better readable using match instead of if-else chains

src/bloqade/jeff/parse/lowering.py
nathanieltornow: this needs to stay a function, since JeffType isn't hashable. however, i changed it to a match-chain to be more readable

src/bloqade/jeff/parse/load.py
nathanieltornow: done via a semantic check

src/bloqade/jeff/parse/load.py
```diff
@@ -0,0 +1,62 @@
...
+                f"{type(source).__name__}"
+            )
+    version = module.version
+    major, minor = jf.schema.schemaVersionMajor, jf.schema.schemaVersionMinor
+    # Semantic versioning: below 1.0 a minor release can break the format. From 1.0
+    # on, a reader also reads each earlier minor release of its major version.
+    if (
+        version is None
+        or version.major != major
+        or (version.minor != minor if major == 0 else version.minor > minor)
```
david-pl: Shouldn't a minor release also work the other way around? E.g., a `v1.1` reader should be able to read `v1.2`, otherwise they'd need to bump it to `v2.0` to mark it breaking.

david-pl (approved):

